Issue #17, opened 2019-10-29
Readme instructions results in Box 'arkcase/arkcase-ce' could not be found


toddkaufmann:
Following the `Vagrantfile` given in Readme.md results in

    $  vagrant up
    Bringing machine 'default' up with 'virtualbox' provider...
    ==> default: Box 'arkcase/arkcase-ce' could not be found. Attempting to find and install...
	default: Box Provider: virtualbox
	default: Box Version: 3.3.1-r1-a
    The box 'arkcase/arkcase-ce' could not be found or
    could not be accessed in the remote catalog. If this is a private
    box on HashiCorp's Atlas, please verify you're logged in via
    `vagrant login`. Also, please double-check the name. The expanded
    URL and error message are shown below:

    URL: ["https://atlas.hashicorp.com/arkcase/arkcase-ce"]
    Error: The requested URL returned error: 404 Not Found

Apparently `vagrant` looks elsewhere for boxes now?  
This can be fixed by supplying a box url as such:

    config.vm.box_url = "https://app.vagrantup.com/arkcase/boxes/arkcase-ce"  config.vm.box_url = "https://app.vagrantup.com/arkcase/boxes/arkcase-ce"

toddkaufmann:
Note this problem will only occur if you don't already have a box called `arkcase/arkcase-ce`
(you can use `vagrant box list` to view existing boxes), 
and seems to be related to vagrant version 
(eg I saw it on v.1.9.5 on Linux, but not OS X v2.2.4).

In either case, adding `box_url` doesn't hurt.

PR #18, opened 2019-10-29
Specify box url in Vagrantfile example

toddkaufmann:
Speciy `config.vm.box_url` specification -- fix for #17


david-oc-miller:
Good stuff @toddkaufmann , and thanks for fixing this.  Definitely it didn't used to be a problem, so somehow the default URL changed in Vagrant.  Anyway, better to be explicit about these things.  Thanks again.

PR #22, opened 2019-11-05
ARKAI-29 better support for FOIA portal directory

david-oc-miller:
Move the LDAP config tasks to their own role, and call that role from ArkCase and FOIA roles


bojmil-armedia (approved):
Looks good.

PR #24, opened 2019-11-07
ARKAI-31 external portal auth fixes

david-oc-miller:


vagrant/provisioning/roles/arkcase-ldap-config/templates/spring-config-ldap.properties
```diff
@@ -32,8 +32,8 @@ ldapConfig.changedGroupSearchFilter=(&(objectclass=group)(modifyTimestamp>=%s))
 ldapConfig.groupSearchPageFilter=(&(objectclass=group)(cn>=%s))
 ldapConfig.changedGroupSearchPageFilter=(&(objectclass=group)(cn>=%s)(modifyTimestamp>=%s))
 ldapConfig.groupsSortingAttribute=cn
-ldapConfig.userPrefix={{ ldap_prefix }}
-ldapConfig.groupPrefix={{ ldap_prefix }}
+ldapConfig.userPrefix={{ ldap_prefix }}{{ portal_prefix }}
+ldapConfig.groupPrefix={{ ldap_prefix }}{{ portal_prefix }}
```
bojmil-armedia: The portal_prefix for the groups should be empty string. So, I think we should have separate variables for user and groups. We do have separate properties as well. So we should have:

- ldap_user_prefix
- ldap_group_prefix
- ldap_portal_user_prefix
- ldap_portal_group_prefix

bojmil-armedia (changes requested):
Please add separate variables for users and groups to support different prefixes.

vagrant/provisioning/roles/arkcase-ldap-config/templates/spring-config-ldap.properties
sasko-tanaskoski: Probably is better this way.

david-oc-miller: I just won't include '{{ portal_prefix }}' on the ldapConfig.groupPrefix; then the group prefix will never have a portal suffix, and no chance of group prefix and user prefix being out of line.  Way too confusing to have different user prefix and group prefix in the same environment.

david-oc-miller: @bojmil-armedia , @sasko-tanaskoski, change applied

sasko-tanaskoski: I'm good with this.

PR #34, opened 2019-12-04
ARKAI-22 pentaho EE use expect instead of echo

david-oc-miller:


david-oc-miller:
finally figured out how expect works

PR #42, opened 2020-01-05
Developer image support

david-oc-miller:
Fixes for Alfresco CE (to use Java 8 versus Java 11); refactor the foia block in the httpd config

### DIFF
--- a/vagrant/provisioning/arkcase-dev-facts.yml
+++ b/vagrant/provisioning/arkcase-dev-facts.yml
@@ -1,0 +1,321 @@
+#################################
+## host-specific settings
+#################################
+email_service_account: "nosuchmailbox@arkcase.com"
+email_service_account_escaped: "{{ email_service_account | replace('@', '%40') }}"
+email_service_password: "not-a-real-password"
+
+calendar_enabled_for_case_files: false
+calendar_purge_option: "RETAIN_INDEFINITELY"
+calendar_enabled_for_complaints: false
+
+# external_host: the host name a browser will use
+external_host: acm-arkcase
+# internal host: the actual host name of the VM
+base_domain: acm-arkcase
+internal_host: "{{ base_domain }}"
+
+activemq_host: "{{ internal_host }}"
+alfresco_host: "{{ internal_host }}"
+arkcase_host: "arkcase-host"
+database_host: "{{ internal_host }}"
+mariadb_host: "{{ internal_host }}"
+pentaho_host: "{{ internal_host }}"
+snowbound_host: "{{ internal_host }}"
+solr_host: "{{ internal_host }}"
+
+
+### ensure samba domain is based on the internal_host
+samba_domain: "DC.ARKCASE-CE"
+
+# server_admin goes in the httpd conf file, and is the email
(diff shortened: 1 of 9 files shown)


vagrant/provisioning/arkcase-dev-facts.yml
```diff
@@ -0,0 +1,321 @@
+#################################
+## host-specific settings
+#################################
```
david-oc-miller, line 3: keep everything just like it is in the old Vagrant VM, except for what must be changed due to using MariaDB (instead of MySQL) and Samba (instead of ApacheDS)

vagrant/provisioning/roles/arkcase-developer/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+- name: create arkcase database schema
+  include_tasks: "{{ role_path }}/../mariadb/tasks/create_schema.yml"
+  args:
+    apply:
+      vars:
+        item: arkcase
```
david-oc-miller, line 6: Since we don't deploy ArkCase itself, we still have to create the database schema so ArkCase's Liquibase can do its thing.

Also, we don't deploy any Pentaho reports yet.  Still need to think through how to do this, since currently we deploy the reports from the arkcase.war.

vagrant/provisioning/roles/httpd/templates/httpd-ssl.conf
```diff
@@ -305,7 +305,8 @@ ProxyPassReverse /share/res https://{{ alfresco_host }}:7070
 ProxyPass /pentaho https://{{ pentaho_host }}:2002/pentaho
 ProxyPassReverse /pentaho https://{{ pentaho_host }}:2002/pentaho
 
-ProxyPassMatch ^/arkcase/stomp/(.*)/websocket$ wss://{{ arkcase_host }}:8843/arkcase/stomp/$1/websocket
+ProxyPass /arkcase/stomp wss://{{ arkcase_host }}:8843/arkcase/stomp
```
david-oc-miller, line 308: This is the fix for the ActiveMQ STOMP issue I had a few months back, the last time I tried to make a Hyper-V VM.  This seems to work fine.  I don't know who came up with the old, weird way.  This way seems to be fine.

vagrant/provisioning/roles/arkcase-developer/tasks/main.yml
bojmil-armedia: We can deploy the Pentaho reports at Arkcase startup, if they are not deployed. Also we can automatically update older reports and that would solve a lot of our issues.

bojmil-armedia (approved):

vagrant/provisioning/roles/arkcase-developer/tasks/main.yml
david-oc-miller: I thought about that... it's really the only way to know we are deploying the right reports.  We can scan for all reports in some certain folder, both in arkcase.war and in the extension jar.  We could send a JMS message, so the report deployment doesn't slow down the startup.

PR #46, opened 2020-01-23
Update server.yml

bojmil-armedia:


vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -51,7 +51,6 @@ application:
     historyDays: 30
     baseUrl: "https://{{ external_host }}/arkcase"
     applicationName: "arkcase"
-    helpUrl: "https://support.arkcase.com/hc/en-us/sections/360001477151-User-Guide"
```
david-oc-miller, line 54: good catch here too, thanks.

PR #47, opened 2020-01-24
Vagrant box build fix

bojmil-armedia:


vagrant/Vagrantfile
```diff
@@ -73,14 +73,14 @@ Vagrant.configure("2") do |config|
...
   # copy the provisioning folder so we can run ansible locally
-  config.vm.provision "file", source: "provisioning", destination: "/tmp/provisioning"
+  config.vm.provision "file", source: "provisioning", destination: "provisioning"
 
   # Enable provisioning with a shell script. Additional provisioners such as
   # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
   # documentation for more information about their specific syntax and use.
   $script = <<-SCRIPT 
     sudo yum -y install ansible
-    cd /tmp/provisioning
```
david-oc-miller, line 83: Instead of this approach, which will leave just the current version in /home/vagrant/provisioning, how about cloning the entire repository?  In the arkcase-developer play, or even here in the Vagrantfile you could "git clone https://github.com/ArkCase/arkcase-ce".

david-oc-miller: Then we could leave the current /tmp/provisioning as is... since often we are testing changes to the build, so we don't always want to build from a newly-cloned repository.  Or anyway, we could copy the current working folder somewhere, and run ansible from that folder; and we can also clone the repository to some other folder, which will support 'git pull' and updating the box later on according to the latest installer changes.

bojmil-armedia: I don't know what was happening with /tmp/provisioning, that's why I change it to the vagrant user home directory.  
As far as updates go, we should have a chat. If developers are building the box (like now), they can pull the latest code and run "vagrant up --provision", which should copy the latest code and run the playbook again.

david-oc-miller:
@bojmil-armedia, why did we need to add the FOIA report definition files to the installer?  What was wrong with the old way, where the installer pulls the report definition files from the arkcase.war file?

bojmil-armedia:
The issue was for the developers box. There we don't have the arkcase.war file. So, the reports needs to be installed from the report definition folder.  
These roles, arkcase-dev-reports and foia-dev-reports, are only used in the arkcase-dev.yml, i.e. only for the dev build.

By the way, the reports definitions for Arkcase core are not extracted from the war file. Only the extension reports are extracted. We probably need to implement this for arkcase.war.

PR #56, opened 2020-02-26
RHEL / FIPS fixes

david-oc-miller:
still need to update Yarn CLI to use sha256 for hashing, versus md5

### DIFF
--- a/vagrant/provisioning/roles/common/tasks/main.yml
+++ b/vagrant/provisioning/roles/common/tasks/main.yml
@@ -26,6 +26,7 @@
       - tar
       - wget
       - sshpass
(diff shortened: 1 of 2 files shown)


vagrant/provisioning/roles/common/tasks/main.yml
```diff
@@ -26,6 +26,7 @@
       - tar
       - wget
       - sshpass
+      - firewalld
```
david-oc-miller, line 29: The FIPS-enabled RHEL 7.5 VMs John created for EEOC did not have these firewall packages

bojmil-armedia (approved):

PR #57, opened 2020-02-27
Snowbound upgrade

bojmil-armedia:


vagrant/provisioning/roles/snowbound-upgrade/tasks/main.yml
```diff
@@ -0,0 +1,107 @@
...
+
+- name: see if war files exist
+  become: yes
+  become_user: snowbound
+  stat:
+    path: "{{ root_folder }}/app/snowbound/webapps/VirtualViewerJavaHTML5.war"
+  register: war_file
+
+- name: Delete all but last 5 backups
+  become: yes
```
david-oc-miller: very nicely done, but assuming much of this code is from the original snowbound role, let's extract it into a common area, like a snowbound-app role or a snowbound-install role that you could call both from snowbound and from snowbound-upgrade... most of this ansible structure I came up with in this project is just from constant application of "don't repeat yourself"

bojmil-armedia: Agree.

PR #60, opened 2020-03-06
pentaho_ee_license_year is updated to the new 2020 licenses

dzdimov:
pentaho_ee_license_year is updated to match the new license zip file


david-oc-miller:
The PR title should be more descriptive, like in this case, your comment should have been the title, so anybody looking at the PR list can see what this one is for...

PR #65, opened 2020-03-23
update the main.yml

dzdimov:
Change to the ansible role in which we copy the -runtime files from the branding folder from the pre-update .arkcase configuration folder to the new version of .arkcase thus retaining the logos and brandings


bojmil-armedia (approved):

david-oc-miller:
@dzdimov  please make the commit messages more descriptive ... "update the main.yml" is not very helpful when this repository probably has three dozen main.yml files.

PR #66, opened 2020-03-31
Adapt Ansible scripts to Docker

fabricetriboix:
This especially includes:
 - Turning off the firewall
 - No SELinux
 - No direct manipulations of host name and /etc/hosts


vagrant/provisioning/roles/mariadb/tasks/main.yml
```diff
@@ -164,5 +164,3 @@
   systemd:
     name: mariadb
     state: restarted
```
david-oc-miller, line 166: Why do we need to always restart MariaDB, instead of restarting only when its configuration has changed?

docker/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.7"
+
+# NB: You must have manually installed ArkCase inside a Docker
```
david-oc-miller, line 3: would you put a Dockerfile into this git repository, or would it be stored somewhere else?  How would somebody go about manually installing ArkCase inside a Docker container?

vagrant/provisioning/roles/mariadb/tasks/main.yml
fabricetriboix: Sorry, I didn't get a notification... That's because on Docker, for some reason, MariaDB is not running by default, so I thought we could restart it in any case. It doesn't hurt, just costs a few seconds and that will work in all situations.

docker/docker-compose.yml
fabricetriboix: As I documented in Confluence, I can't write a Dockerfile. The reason is that Docker just allows basic operations while building the image, and especially doesn't run any daemons. But the Ansible scripts expect systemd to be running, along with a number of services.

So for the time being, we have to build the Docker image manually until I have extracted enough services from it to allow a simpler process. All the orchestration between services will be done by the IaC/deployment tool, such as CloudFormation, instead of Ansible scripts. The Ansible scripts will be limited to deploying their respective software and won't do any cross-service jobs.

david-oc-miller: OK... how would somebody build the docker image manually?

fabricetriboix: There are step-by-step instructions there: https://arkcase.atlassian.net/wiki/spaces/DO/pages/6422529/Notes+about+Dockerizing+ArkCase

david-oc-miller: @fabricetriboix if you add this link into the docker-compose.yml then we can merge this one.

docker/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+# container and then committed in into an image like so:
+# `$ docker commit MYCONTAINER arkcase-ce-core:3.3.5`
+#
+# For more details, check
+# [this](https://arkcase.atlassian.net/wiki/spaces/DO/pages/6422529/Notes+about+Dockerizing+ArkCase).
```
david-oc-miller, line 8: :-)

david-oc-miller (dismissed):

david-oc-miller (approved):

kemboi22:
https://arkcase.atlassian.net/wiki/spaces/DO/pages/6422529/Notes+about+Dockerizing+ArkCase

this is link  not working

PR #69, opened 2020-04-07
New role for starting config server

kikomanaskov333:
The new ansible role is performing the following tasks:
- Start config server
- Wait for config server
- Start arkcase


bojmil-armedia:
Let's discuss where this role should be used. We should add it in some playbook.

dzdimov:
We should probably use this role in the ArkCase CE installation since ArkCase and config server are not started by default there. @david-oc-miller has more explanation on this

david-oc-miller:
@bojmil-armedia , it's on purpose that I don't start config server and ArkCase for CE; because the purpose of the CE build is to package a Vagrant box, and the logs being generated by ArkCase inhibit creation of the box.  I had some file corruption errors on starting the box from the image, and tracked it down to the acmScheduledTasks.json file actually; sometimes the image was taken exactly when acmScheduledTasks.json was at 0 length, preparatory to updating it with new values, and the 0-length acmScheduledTasks.json kept ArkCase from starting.

Hence not starting ArkCase as part of the CE play.

It makes sense to start ArkCase as part of arkcase-dev of course.

vagrant/provisioning/roles/arkcase-post-upgrade-start-config/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+- name: start config server
+  become: yes
+  systemd:
+    name: "config-server"
+    state: started
+
+- name: wait for config server
```
david-oc-miller: I don't understand the name of this role, since the upgrade plays already start both config server and arkcase, see roles/arkcase-post-upgrade/tasks/main.yml.  

Also you should remove this code from roles/arkcase-post-upgrade/tasks/main.yml, so as to avoid code duplication, and then call this role from all the plays where you want to start ArkCase after initial installation (arkcase-dev.yml, all the various arkcase-ee-... roles, basically any play except for arkcase-ce.yml.

david-oc-miller: Also of course add this role to the various upgrade plays, since you will have removed it from roles/arkcase-post-upgrade/tasks/main.yml.

david-oc-miller: And consider renaming the role itself to something like 'start-arkcase', which describes what it does, versus when it is called.

vagrant/provisioning/arkcase-ce-external-ldap-foia-pki.yml
```diff
@@ -20,6 +20,7 @@
     - pentaho-license
     - arkcase-prerequisites
     - arkcase-app
+    - start-arkcase
```
david-oc-miller: foia role must run first before arkcase can be started, may want to always put start-arkcase as the very last role

david-oc-miller: @kikomanaskov333 I guess you're going to push another commit restoring the changes you reverted, with start-arkcase at the end of each play?

kikomanaskov333: Yes @david-oc-miller , now i'm going to push right commit with corrected plays.

bojmil-armedia (approved):

dzdimov (approved):

PR #70, opened 2020-04-08
use pre-existing alfresco

david-oc-miller:
The point of this merge request is to allow us to install ArkCase into an environment that already has Alfresco.  In this case we only have to run the new alfresco-site play, so as to create the acm site.


vagrant/provisioning/roles/alfresco-site/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+    body_format: json
+    body:
+      name: "{{ item }}"
+      title: "{{ item }}"
+  register: folder_create_out
+  failed_when: '"site" not in folder_create_out.content'
+  loop: "{{ content_management_folders }}"
+  when: 'item not in existing_folders_out.content'
+
+- name: create records management share site
```
bojmil-armedia, line 102: We can introduce a variable not to install the RM site, when clients don't want it.

bojmil-armedia (approved):

vagrant/provisioning/roles/alfresco-site/tasks/main.yml
david-oc-miller: @bojmil-armedia we have 'alfresco_install_governance_services' in the facts file already ... but in the alfresco role we may or may not check it.

PR #71, opened 2020-04-10
Arkcase.next installer (confluent-platform)

kikomanaskov333:
Added role that support installation of confluent-platform.


vagrant/provisioning/arkcase-app-upgrade.yml
```diff
@@ -13,5 +13,7 @@
       tags: [foia]
     - role: arkcase-post-upgrade
       tags: [core]
+    - role: start-arkcase
```
david-oc-miller: I see we are mixing up changes from another ticket in this pull request.  Also, 'start-arkcase' should always be last.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: "{{ packages }}"
+  vars:
+    packages:
+    - curl
+    - which
+
+- name: Install Confluent Platform public key
+  rpm_key:
+    state: present
+    key: https://packages.confluent.io/rpm/5.4/archive.key
```
david-oc-miller: Let's pull the version (5.4) into a fact, so we can upgrade or change version just by changing the facts file.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+- name: Create Confluent repo (1/2)
+  yum_repository:
+    name: Confluent.dist
+    description: Confluent repository (dist)
+    baseurl: https://packages.confluent.io/rpm/5.4/7
```
david-oc-miller: I suppose '7' is the minor version, let's extract this into the facts file too.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+- name: Clean the yum caches
+  command: yum clean all
+  args:
+    warn: False
+
+- name: Install Confluent Platform
+  yum:
+    name: confluent-platform-2.12
+    state: present
+
```
david-oc-miller: '2.12' is also ripe for extraction to the facts file.

We would like for Kafka to put its application (like log4j log files) logs into '{{root_folder}}/log/kafka', which according to https://serverfault.com/questions/831951/kafka-logfile-location you may be able to do by setting LOG_DIR environment variable in the SystemD unit file.  Please see if you can do this.

We would like Kafka to store its topic data in {{root_folder}}/arkcase/data/kafka, which according to https://docs.confluent.io/current/kafka/deployment.html may be done by setting logs.dir in config/server.properties.  Try and get this to work too.

Does installing confluent-platform-2.12 also enable the SystemD services and start the services?  If not you may need to add these steps.

Please check with Nadica Deleva on configuring the JMS adapter and the JDBC adapter.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
+---
+# tasks file for confluent-install
+- name: Install curl and which
+  yum:
+    name: "{{ packages }}"
+  vars:
```
david-oc-miller, line 6: don't forget to set default values of the new facts here...

bojmil-armedia:
Someone from the Arkcase.Next team should work with @kikomanaskov333,  run this locally and verify if everything needed is installed and correctly configured.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
kikomanaskov333: @david-oc-miller with the last commit i've already defined the variables and their default values in the following facts file: arkcase-ce-facts.yml and arkcase-dev-facts.yml.
I hope it's a proper way, please check

david-oc-miller: @kikomanaskov333 there are many, many other facts files in our Ansible Tower server (one per environment).  To keep from having to update all these other facts files (which do not live here in GitHub) we need to define default values here.  Ansible has a default filter you can use like so: {{ confluent_version | default("5.2") }}, that should do it.  See here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#defaulting-undefined-variables

vagrant/provisioning/arkcase-app-upgrade.yml
david-oc-miller: you should remove these commits or revert these changes from this pull request, since you already applied these changes in the other pull request.

vagrant/provisioning/roles/confluent-platform-install/templates/confluent-kafka.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Apache Kafka - broker
+Documentation=http://docs.confluent.io/
+After=network.target confluent-zookeeper.target
+
+[Service]
+Type=simple
+User=cp-kafka
+Group=confluent
+Environment="LOG_DIR=/opt/arkcase/log/kafka"
```
david-oc-miller: Here you need to use the variable: LOG_DIR={{root_folder}}/log/kafka

vagrant/provisioning/roles/confluent-platform-install/templates/confluent-kafka.service
```diff
@@ -0,0 +1,18 @@
...
+ExecStart=/usr/bin/kafka-server-start /etc/kafka/server.properties
+ExecStop=/opt/kafka/bin/kafka-server-stop.sh
```
david-oc-miller, line 12: This should cause the Kafka application log files to be written to {{root_folder}}/log/kafka, but we still need configuration to cause the data files to go into {{root_folder}}/data/kafka.  By default they will go into /tmp which definitely we don't ant.

bojmil-armedia:
@kikomanaskov333 we should make sure all services used are using TLS. All other services (except Config Server) already have this implemented. If the certificates are signed by the same CA, we can use them immediately with our current java truststore.  
From the services I'm aware of we need TSL in Kafka, Schema Registy, Confluent Control center, Zookeeper.

david-oc-miller:
@kikomanaskov333, everything looks good.  One thing is missing, which is to configure Kafka so it stores data files under {{ root_folder }}/data/kafka.

kikomanaskov333:
> @kikomanaskov333, everything looks good. One thing is missing, which is to configure Kafka so it stores data files under {{ root_folder }}/data/kafka.

kikomanaskov333:
@david-oc-miller , now i'm configuring Kafka to store its data files under {{ root_folder }}/data/kafka . Instead of data files, there are many Kafka config files which is stored under /etc/kafka/ by default. Does this mean that all configs also need to be stored under {{ root_folder }}/data/kafka ?
I did a mistake by inadvertent clicking close pull request, and that's why it is reopen

david-oc-miller:
@kikomanaskov333 , since we install Kafka from RPM packages, we don't have to put the configuration under {{root_folder}}/app/kafka; we don't do that for MariaDB or httpd.  For httpd and MariaDB we leave the configuration files where the RPM package puts them; but we do alter the config files so the logs and data files go where we want.  For Kafka you can do the same.

vagrant/provisioning/arkcase-app-upgrade.yml
david-oc-miller: @kikomanaskov333 it seems like you still have to fix or remove these `start-arkcase` lines

vagrant/provisioning/confluent-platform.yml
```diff
@@ -0,0 +1,3 @@
+- hosts: localhost
```
david-oc-miller: @kikomanaskov333 and add `confluent-platform` to the arkcase-ce.yml and arkcase-dev.yml playbooks

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    owner: cp-kafka
+  loop:
+    - "{{ root_folder }}/data/kafka"
+    - "{{ root_folder }}/log/kafka"
+
+- name: Change kafka data location
+  become: yes
+  template:
+    src: server.properties
+    dest: /etc/kafka/
```
david-oc-miller, line 62: ask for a backup of this file...

vagrant/provisioning/arkcase-app-upgrade.yml
kikomanaskov333: @david-oc-miller, I will remove these commits from this pull request

vagrant/provisioning/roles/start-arkcase/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+- name: start config server
+  become: yes
+  systemd:
+    name: "config-server"
+    state: started
+
+- name: wait for config server
+  wait_for:
+    connect_timeout: 15
+    delay: 5
```
david-oc-miller: maybe if you `git pull origin develop` it will resolve these changes that you already made in the earlier PR

kikomanaskov333:
@david-oc-miller, i think the whole installation is fine. I tested on my own environment and everything was ok. It remains to be tested with a real configuration.

vagrant/provisioning/roles/confluent-platform-install/templates/server.properties
```diff
@@ -28,7 +28,7 @@ broker.id=0
 #     listeners = listener_name://host_name:port
 #   EXAMPLE:
 #     listeners = PLAINTEXT://your.host.name:9092
-listeners=PLAINTEXT://:9093
+listeners=PLAINTEXT://:{{ kafka_port | default(9093) }}
```
david-oc-miller, line 31: :-)

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+- name: Reload daemon files if needed
+  become: yes
+  command: systemctl daemon-reload
+  when: systemd_kafka_updated is changed
+
+- name: Enable zookeeper to start on boot
+  become: yes
+  systemd:
+    name: confluent-zookeeper
+    enabled: yes
```
david-oc-miller, line 83: @kikomanaskov333 , one last change is to leave Confluent disabled by default, so a specific installation can opt-in to it.  Since most sites won't be using Kafka yet.

```yaml
- name: Enable zookeeper to start on boot
  become: yes
  systemd:
    name: confluent-zookeeper
    enabled: "{{ enable_kafka | default(false) }}"
...
```

And same for confluent-kafka.

david-oc-miller: everything looks good in my local VM.

dzdimov (dismissed):
I think that everything looks good except that we also might need a enable_kafka fact added

kikomanaskov333:
@david-oc-miller , i think this could be a good approach. The services, kafka and zookeeper, will be enabled and started only if variable enable_kafka is defined and has a value 'yes' otherwise there will be installation of the platform without starting services

david-oc-miller:
@kikomanaskov333 thanks, it works well; the services stay disabled by default, and when I set enable_kafka=yes, they get enabled.

david-oc-miller (approved):

PR #73, opened 2020-04-14
Fixed ldap user and group prefixes with numbers

bojmil-armedia:


NCuculova (approved):

vagrant/provisioning/roles/arkcase-ldap-config/templates/ldap.yaml
```diff
@@ -38,11 +38,10 @@
         unicodePwd: unicodePwd
         givenName: givenName
         objectClass: top,person,user,organizationalPerson
-        id: arkcase
         cn: cn
         sn: sn
         userPrincipalName: userPrincipalName
-        userAccountControl: userAccountControl
+        userAccountControl: 512
```
david-oc-miller, line 44: I'm curious why we have a specific value here (512) but the other properties have property names, instead of constant values?

bojmil-armedia: 512 is a NORMAL_ACCOUNT in ActiveDirectory, Some directories might want to change this value though.  
The other values reference field names in the application AcmUser and some special values set in MapperUtils ("unicodePwd" for example).

PR #74, opened 2020-04-17
Add tags to all plays, so we can run one specific role

david-oc-miller:


vagrant/provisioning/roles/ldap/tasks/add-user.yml
```diff
@@ -37,6 +37,9 @@
           replace: unicodePwd
           unicodePwd:: {{ base64_password.stdout }}
         dest: "{{ root_folder }}/install/password_change.ldif"
+    - name: wait a few seconds
```
david-oc-miller, line 40: With AD it takes a few seconds for the new user to actually be visible

david-oc-miller:
I'm gonna merge this with no review since I want to use it to push some ldap changes to EEOC via AWX

bojmil-armedia:
This is a good idea and it looks good.

PR #75, opened 2020-04-18
Wait for Alfresco exploded webapp to be available

david-oc-miller:
Ran into this one when building a new box locally.

### DIFF
--- a/vagrant/provisioning/roles/alfresco/tasks/main.yml
+++ b/vagrant/provisioning/roles/alfresco/tasks/main.yml
@@ -470,6 +470,11 @@
     delay: 3
     timeout: 180
 
+- name: now wait for Alfresco to really become available
+  wait_for:
+    path: "{{ root_folder }}/app/alfresco/webapps/alfresco/WEB-INF/web.xml"
+    delay: 5
+    timeout: 180
 
 # make it so the filter URL pattern for the sync service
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -470,6 +470,11 @@
     delay: 3
     timeout: 180
 
+- name: now wait for Alfresco to really become available
+  wait_for:
+    path: "{{ root_folder }}/app/alfresco/webapps/alfresco/WEB-INF/web.xml"
+    delay: 5
+    timeout: 180
```
bojmil-armedia: This file might exist if this is not the first Alfresco startup. I guess it is ok in this script.  
Anyway, we can also use uri. Here is an example:
```
    - name: Wait for URI
      uri:
        url: https://someURI
        method: GET
        user: admin
        password: password123
        force_basic_auth: yes
        status_code: [201, 202]
        return_content: yes
        body_format: json
      register: result
      until: result.json.success == true
      retries: 180
      delay: 30
```

david-oc-miller: interesting idea... let me try it.

kikomanaskov333 (approved):

bojmil-armedia (approved):

PR #76, opened 2020-04-21
Separate the FOIA portal version from the core ArkCase version

bojmil-armedia:


dzdimov (approved):

david-oc-miller:
I'm OK with the change, but what was the use case for it?

david-oc-miller (approved):

bojmil-armedia:
EEOC is deploying Arkcase version 3.3.7-RC2 and the external portal version eeoc-fts-3.3.7-RC3

PR #79, opened 2020-04-27
ACP-32 Installer support for elasticsearch

kikomanaskov333:
This ansible role 'elasticsearch-install' is doing installation and basic configuration for ElasticSearch service


dzdimov (dismissed):

bojmil-armedia:
The role name should be just **elasticsearch**. That's the naming convention we have until now.  
Perhaps, it does make sense to have multiple roles: install, configure, upgrade. It is more granular and we can pick the roles to execute.

vagrant/provisioning/roles/elasticsearch-install/templates/elasticsearch.yml
```diff
@@ -0,0 +1,14 @@
+cluster.name: {{ cluster_name | default('acm-cluster') }}
+node.name: {{ node_name | default('acm-node01') }}
+path.data: {{ root_folder }}/data/elasticsearch
+path.logs: {{ root_folder }}/log/elasticsearch
+#path.conf: /etc/elasticsearch
+network.host: localhost
+http.port: {{ ES_PORT | default(9200) }}
```
bojmil-armedia: We don't use upper case naming of variables. I would name the variable with longer name **elasticsearch_port**.

kikomanaskov333: @bojmil-armedia , since this role is only for installation and basic configuration of elasticsearch service, i think it would be nice if it remains with the name elasticsearch-install. Also, i guess there will be another task for configuration which will be wrapped into the new role 'elasticsearch-configure'. For me, there is no problem to change it

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+        - "{{ root_folder }}/data/elasticsearch"
+        - "{{ root_folder }}/log/elasticsearch"
+
+    - name: Copying ElasticSearch configuration files
+      template:
+        src: elasticsearch.yml
+        dest: /etc/elasticsearch
+        backup: yes
+    
+    - name: Adding services to auto start for CentOS 6
```
david-oc-miller: we don't need to support CentOS 6

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+        name: elasticsearch
+        enabled: yes
+        daemon_reload: yes
+        masked: no
+        state: started
+      when: ansible_distribution == 'CentOS' and ansible_distribution_major_version == '7' and enable_elasticsearch is defined and enable_elasticsearch == "yes"
+ 
+    - name: Wait for 20 seconds for elasticsearch to become responsive...
+      wait_for: timeout=20
+      delegate_to: localhost
```
david-oc-miller: When running this role on the AWX server, will this 'localhost' be the AWX server itself, or the remote host?

vagrant/provisioning/roles/elasticsearch-install/templates/elasticsearch.yml
```diff
@@ -0,0 +1,14 @@
+cluster.name: {{ cluster_name | default('acm-cluster') }}
```
david-oc-miller: @bojmil-armedia , @fabricetriboix , do we prefer generic 'cluster_name' and 'node_name' here, or would we rather have 'elastic_cluster_name' and 'elastic_node_name'

bojmil-armedia: It is more descriptive and specific with **elastic** prefix. Or **elasticsearch** as the other variables.

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
kikomanaskov333: @david-oc-miller, since 'localhost' will be the AWX server itself, i removed 'delegate_to'. Now this part of the task will be running on remote host.
Dimitar told me that 'delegate_to: localhost' was using when you didn't have AWX server.

dzdimov (dismissed):

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Adding services to auto start
+      systemd:
+        name: elasticsearch
+        enabled: yes
+        daemon_reload: yes
+        masked: no
+        state: started
+      when: enable_elasticsearch is defined and enable_elasticsearch == "yes"
+ 
+    - name: Wait for 30 seconds for elasticsearch to become responsive...
```
david-oc-miller: Here consider waiting for the elasticsearch webapp to be responsive, like with Bojan's comment here, https://github.com/ArkCase/arkcase-ce/pull/75

kikomanaskov333: @david-oc-miller , in my last commit I included the 'uri' ansible module with the main role of waiting for elasticsearch to be responsive. In a consultation with Dimitar, as a variable for remote host where elasticsearch will be up and running, i use 'external_host'.

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+        name: elasticsearch
+        enabled: yes
+        daemon_reload: yes
+        masked: no
+        state: started
+      when: enable_elasticsearch is defined and enable_elasticsearch == "yes"
+ 
+    - name: Wait for elasticsearch to become responsive...
+      uri:
+        url: "{{ external_host | default('http://localhost') }}:{{ elasticsearch_port | default(9200) }}"
```
david-oc-miller: For a single-VM install where httpd has already started using `external_host` probably works, but for AWS or other sites where the httpd/haproxy may not be in place yet, this might not work.  You might want to check localhost after all.

Also, consider including `validate_certs: no` to the parameters, in case the elastic host is on a self-signed cert.

Also, add the `when` from line 43 to this task as well, so you only wait for Elastic to start when elastic is enabled, and you don't wait for it to start when Elastic is disabled.

david-oc-miller (approved):

PR #83, opened 2020-05-05
Require admin access for Solr web application

kikomanaskov333:
Additions to the 'httpd role' are added in order to achieve only LDAP authenticated users to have access to Solr web app.

### DIFF
--- a/vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-linuxadmins-helper.script2
+++ b/vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-linuxadmins-helper.script2
@@ -1,0 +1,14 @@
+AuthType Basic
+AuthName "Authentication Required"
+AuthBasicProvider ldap
+RewriteEngine On
+#LDAPReferrals Off
+# Perform LDAP lookup for user to check group membership
+AuthLDAPUrl "{{ ldap_url }}/{{ ldap_base }}?{{ ldap_remote_user_attribute | default('sAMAccountName') }}?sub?(objectClass=*)"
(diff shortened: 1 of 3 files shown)


vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-linuxadmins-helper.script2
```diff
@@ -0,0 +1,14 @@
...
+RewriteEngine On
+#LDAPReferrals Off
+# Perform LDAP lookup for user to check group membership
+AuthLDAPUrl "{{ ldap_url }}/{{ ldap_base }}?{{ ldap_remote_user_attribute | default('sAMAccountName') }}?sub?(objectClass=*)"
+AuthLDAPBindDN "{{ ldap_base }}"
+AuthLDAPBindPassword "{{ ldap_bind_password }}"
+AuthzLDAPAuthoritative off
+#AuthLDAPInitialBindAsUser off
+AuthLDAPRemoteUserAttribute "{{ ldap_remote_user_attribute | default('sAMAccountName') }}"
+Require ldap-filter &(objectClass=person)(objectClass=user)(memberOf=CN="{{ ldap_CN | default ('linuxadmins') }}","{{ ldap_base }}")((userAccountControl:1.2.840.113556.1.4.803:=2))
```
david-oc-miller: Let's make ldap_CN more descriptive, something like 'solr_web_console_access_group_name'.

Not sure what to do about userAccountControl which is Active Directory specific.  Let's just remove that clause, since if the user's account is deactivated they won't be able to authenticate anyway.

david-oc-miller (approved):

PR #92, opened 2020-05-15
Added improvments in httpd-ssl.conf and ldap-auth-restricted-solr

kikomanaskov333:


dzdimov (dismissed):

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
```diff
@@ -5,10 +5,12 @@ RewriteEngine On
 #LDAPReferrals Off
 # Perform LDAP lookup for user to check group membership
 AuthLDAPUrl "{{ ldap_url }}/{{ ldap_base }}?{{ ldap_remote_user_attribute | default('sAMAccountName') }}?sub?(objectClass=*)"
-AuthLDAPBindDN "{{ ldap_base }}"
+AuthLDAPBindDN "{{ ldap_base | regex_search('(?<=\CN=)(.*?)(?=\,)') }}@{{ ldap_user_domain }}"
```
david-oc-miller: Is AuthLDAPBindDN supposed to be a distinguished name, or a user id?  ldap_base obviously is not a user id at all ("OU=EEOC,OU=Clients,DC=appdev,DC=armedia,DC=com").  Should you be using ldap_bind_user here instead?

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
```diff
@@ -5,10 +5,12 @@ RewriteEngine On
...
 AuthLDAPBindPassword "{{ ldap_bind_password }}"
 #AuthzLDAPAuthoritative off
 #AuthLDAPInitialBindAsUser off
 AuthLDAPRemoteUserAttribute "{{ ldap_remote_user_attribute | default('sAMAccountName') }}"
-Require ldap-filter &(objectClass=person)(objectClass=user)(memberOf="{{ solr_web_console_access_group_name | default(arkcase_admin_group) }}")
-require valid-user
+<RequireAny>
+   Require ip "{{ arkcase_host_address | default('127.0.0.1') }}"
```
david-oc-miller, line 14: Here again we need a larger whitelist.  Operators need access to the Solr console; they would be coming from a VPN address, not from the arkcase host.

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
kikomanaskov333: AuthLDAPBindDN shoud be in the following form for ex. "svc.ark.core.ldap@appdev.armedia.com"...
The first string before @ is the value of CN, and instead of using new fact for CN, i'm using regex for extracting only CN from ldap_bind_user. 
You are right, instead of ldap_base, it should be ldap_bind_user

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
```diff
@@ -5,7 +5,7 @@ RewriteEngine On
 #LDAPReferrals Off
 # Perform LDAP lookup for user to check group membership
 AuthLDAPUrl "{{ ldap_url }}/{{ ldap_base }}?{{ ldap_remote_user_attribute | default('sAMAccountName') }}?sub?(objectClass=*)"
-AuthLDAPBindDN "{{ ldap_base | regex_search('(?<=\CN=)(.*?)(?=\,)') }}@{{ ldap_user_domain }}"
+AuthLDAPBindDN "{{ ldap_bind_user | regex_search('(?<=\CN=)(.*?)(?=\,)') }}@{{ ldap_user_domain }}"
```
david-oc-miller, line 8: this looks OK for the common case; but the fact is that commas are allowed to be in common names ("CN=Miller, David,CN=Users,DC=armedia,DC=com").  Not sure how you would handle this from a regexp perspective.  Anyway for now I guess this will mostly work.

kikomanaskov333: I didn't take it into consideration. The regex searches for the first string between 'CN=' and the first comma to appear.

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
bojmil-armedia: This is only for Arkcase application. It won't require a valid-user, it will allow unauthenticated request from Arkcas instance. All other addresses will require LDAP authentication.

bojmil-armedia: All other addresses will have to authenticate. Only the Arkcase instance is allowed unauthenticated.

bojmil-armedia (approved):

vagrant/provisioning/roles/httpd/templates/ldap-auth-restricted-solr
david-oc-miller: Ah OK, I didn't see the 'RequireAny'.  Thanks.

Does this work even when arkcase_host_address is a host name, versus an IP?

bojmil-armedia: Not sure. This is used only in the arkcase-dev-facts.yaml really, because Arkcase is on a different server.  
What I've read, this could be a space separated list of ip adresses, which should be used in a case when we have multiple Arkcase instances.

david-oc-miller: https://httpd.apache.org/docs/2.4/howto/access.html .... 'require ip' must be an IP address.  'require host' is for a host name.  Also, they are both singular (no space separated list).  'require ip' is "an IP address, a partial IP address, a network/netmask pair, or a network/nnn CIDR specification".  If ArkCase is on multiple hosts you could include one 'require' for each host.

Anyway consider checking by regexp if arkcase_host_address is an IP or a hostname, and issuing 'require ip' or 'require host' depending ... or add some additional facts e.g. arkcase_host_ip_address so it's more clear we must have an IP there.

bojmil-armedia: I have read about the space separated list of IP addresses in the first comment here: https://serverfault.com/questions/815693/allow-multiple-ips-with-the-require-directive-in-apache-2-4
Not sure if it works...

david-oc-miller: Ah nice!

PR #94, opened 2020-05-15
Alfresco request header

kikomanaskov333:


vagrant/provisioning/roles/httpd/templates/httpd-alfresco-redirect.conf
```diff
@@ -280,6 +280,12 @@ SSLCertificateChainFile "{{ ssl_ca }}"
 ProxyPass /alfresco https://{{ internal_host }}:7070/alfresco
 ProxyPassReverse /alfresco https://{{ internal_host }}:7070/alfresco
 
+<LocationMatch "/alfresco">
+  <If "%{REMOTE_ADDR} != '127.0.0.1'">
```
david-oc-miller: The whitelist can't be hardcoded to 127.0.0.1.  On any AWS environment the httpd service will run in a public subnet, and ArkCase and Alfresco will run in private subnets, possibly on different virtual machines.  

You need a fact like 'alfresco_header_auth_whitelist" or something, which can default to 127.0.0.1, but you definitely can't hardcode 127.0.0.1.

bojmil-armedia: This is a whitelist for Arkcase instances. They can use the X-Alfresco-Remote-User header. No one else.  
Anyway this should be **{{ arkcase_host_address | default('127.0.0.1') }}** because the arkcase_host_address for development is on a different address (it's in the arkacase-dev-facts.yaml)

bojmil-armedia (approved):

PR #95, opened 2020-05-15
Added follow_redirects for the uri module of alfresco role since it was failing

dzdimov:
We need to add "follow_redirects: all" since the link is redirecting us and causes the play to fail

### DIFF
--- a/vagrant/provisioning/roles/alfresco/tasks/main.yml
+++ b/vagrant/provisioning/roles/alfresco/tasks/main.yml
@@ -474,6 +474,7 @@
   uri:
     url: https://localhost:7070/share/page/
     validate_certs: false
+    follow_redirects: all
   retries: 10
   delay: 10
 


vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -474,6 +474,7 @@
   uri:
     url: https://localhost:7070/share/page/
     validate_certs: false
+    follow_redirects: all
```
david-oc-miller, line 477: is the installer working now with this change?

bojmil-armedia (approved):

vagrant/provisioning/roles/alfresco/tasks/main.yml
dzdimov: Yes, we tried the change locally on the arkcase-ce aws instance that we are preparing for the MarketPlace

david-oc-miller (approved):

PR #96, opened 2020-05-15
Increased delay time to '60' for uri module of alfresco role since it was falling'

kikomanaskov333:
We need to increase "delay : 60" since we don't receive HTTP 200 as a response for the lower value of the delay.


bojmil-armedia (dismissed):

dzdimov (changes requested):
Let's test this thoroughly on a new instance and see what the delay attribute should be since it proved to have problems with delay: 60 as well

bojmil-armedia (approved):

PR #98, opened 2020-05-20
Efts 436 

dzdimov:
changed the location of extra_artifacts task inside arkcase-app/tasks/main.yml since it's needed to run after the unzip of the configuration files.


bojmil-armedia (approved):

david-oc-miller:
Why do the extra artifacts have to be run after unzipping the config files?

Since the EFTS Jira project is not publicly accessible, please include the ticket title in pull request title, so when we look at a list of pull requests, we can see something sensible, instead of only "Efts 436".

PR #104, opened 2020-06-02
Snowbound 5.2 upgrade

bojmil-armedia:


bojmil-armedia:
Running **snowbound** role with **snowbound_vendor_version: "5.2"** fact will update the JAVA_HOME on existing installations.

vagrant/provisioning/roles/snowbound/tasks/main.yml
```diff
@@ -47,7 +47,7 @@
           catalina_out: "{{ root_folder }}/log/snowbound/catalina.out"
           catalina_temp: "{{ root_folder }}/tmp/snowbound"
           log_folder: "{{ root_folder }}/log/snowbound"          
-          java_home: "/usr/lib/jvm/java-1.8.0-openjdk"
+          java_home: "{{ '/usr/lib/jvm/java-1.8.0-openjdk' if snowbound_vendor_version == '4.1' else '/usr/lib/jvm/java-11-openjdk' }}"
```
david-oc-miller: Is '4.1' == '4.10'?  Otherwise looks great

david-oc-miller: ... or consider version comparison using version() function, https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#version-comparison

bojmil-armedia: It is 4.10 thanks! Updated that in the latest commit.
version() function can be used here. I see we are using it to compare greater than, less than...

david-oc-miller (approved):

PR #106, opened 2020-06-09
Added arkcase-ce-AWS.yml

kikomanaskov333:
New ArkCase-CE playbook file is created with hosts set to all since we need to use for ArkCase CE installation on AWS through AWX


kikomanaskov333:
In a discussion with Dimitar, currently we don't need this new play file. Therefore, I close this pull request.

PR #113, opened 2020-06-23
Arkcase.next

bojmil-armedia:


vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
```diff
@@ -33,15 +33,21 @@
...
+      register: elasticsearch_systemd
+
+    - name: enable ElasticSearch to start on boot
+      become: yes
       systemd:
         name: elasticsearch
         enabled: yes
-        daemon_reload: yes
         masked: no
-        state: started
```
kikomanaskov333, line 42: @bojmil-armedia , if you exclude state: started, than elasticsearch will not start . Only it will be enabled for the next boot.
Also i suggest you add another task before this one for reload daemon if needed since with copying our elastisearch.service template into /etc/systemd/system/elasticsearch.service we did changes and we must reload daemon.
For ex. :
- name: Reload daemon files if needed
  become: yes
  command: systemctl daemon-reload
  when: elasticsearch_systemd is changed

And after that if state is = started then elasticsearch will start running.

vagrant/provisioning/roles/microservices/tasks/main.yaml
```diff
@@ -0,0 +1,26 @@
...
+    state: enabled
+    immediate: yes
+  register: microservices_ports
+  loop:
+    - "9092/tcp"   # Kafka
+
+- name: Reload firewalld
+  become: yes
+  command: firewall-cmd --reload
+  when: microservices_ports is changed
```
david-oc-miller: with `immediate: yes` you don't actually need to call `firewall-cmd --reload`.

vagrant/provisioning/roles/microservices/tasks/main.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+- set_fact:
+    microservices_httpd_proxy_settings: "{{ lookup('template', 'templates/ssl.conf') }}"
+
+- name: Add HTTPD proxy settings for microservices
+  become: yes
+  blockinfile:
+    path: /etc/httpd/conf.d/ssl.conf
+    block: "{{ microservices_httpd_proxy_settings }}"
+    marker: "# {mark} ANSIBLE MANAGED BLOCK for microservices"
```
david-oc-miller, line 17: You can remove the `set_fact` and put the lookup directly into `block`: `block: "{{ lookup('template', 'templates/ssl.conf') }}"`

bojmil-armedia: OK. Thanks.

vagrant/provisioning/roles/microservices/tasks/main.yaml
bojmil-armedia: I have copied this from somewhere. And yes it is not needed.

vagrant/provisioning/roles/elasticsearch-install/tasks/main.yml
bojmil-armedia: Right. Not sure when I deleted this line.

david-oc-miller (approved):

PR #119, opened 2020-07-01
User spreadsheet to installer json

david-oc-miller:
From EEOC we got a user spreadsheet with about 100 users.  This script translates such a spreadsheet to user YAML which we can then plug into an installer configuration file.


vagrant/provisioning/roles/ldap/files/userXlsxToInstallerYaml.py
```diff
@@ -0,0 +1,53 @@
...
+def formatGroupProperty(group):
+    """Format a group name with the proper ldap_prefix and ldap_group_base placeholders"""
+    return '      - "CN={{ ldap_prefix }}' + group[0:group.find('@')] + ',{{ldap_group_base}}"'
+
+for r in range(start_row, sheet.max_row + 1):
+    firstName = getCellValue(sheet, r, 1)
+    lastName = getCellValue(sheet, r, 2)
+    email = getCellValue(sheet, r, 3)
+    groups = getCellValue(sheet, r, 6)
+    office = getCellValue(sheet, r, 8)
```
david-oc-miller, line 33: `office` is EEOC-specific ... when we support another customer this way we may end up making the script more generic.  Or maybe each customer will have a one-off based on this example.

PR #129, opened 2020-07-20
Microservices local fixes

bojmil-armedia:


vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -81,6 +82,24 @@
...
+  changed_when: false
+
+- name: Enable Control Center to work with 1 broker
+  lineinfile:
+    path: /etc/confluent-control-center/control-center-production.properties
+    state: present
+    insertafter: EOF
+    line: "{{ item }}"
+  loop:
+    - "confluent.controlcenter.internal.topics.replication=1"
```
david-oc-miller, line 97: Do we want controlcenter in all environments, or only in local development environments?  Most upper environments will have more than one broker.

bojmil-armedia: We would use control center in all environments. But the current installation cannot add multiple brokers anyway. This will need to be configurable I guess.

PR #131, opened 2020-07-20
added privacy role to arkcase-app-upgrade.yml playbook

bojmil-armedia:


vagrant/provisioning/arkcase-app-upgrade.yml
```diff
@@ -11,6 +11,8 @@
       tags: [core]
     - role: foia
       tags: [foia]
+    - role: privacy
```
david-oc-miller, line 14: I guess we control whether foia or privacy is activated by using tags?

bojmil-armedia: Yes. We trigger the deployment with tags from the gitlab-ci.yaml file on every build.

PR #134, opened 2020-08-05
centos7 hyper-v json fixed with correct syntax and url

marjan4jc:
Current CentOS7 ISO URL (http://mirror.vcu.edu/pub/gnu_linux/centos/7.7.1908/isos/x86_64/CentOS-7-x86_64-Minimal-1908.iso) is not available. I put this one (http://mirror.vcu.edu/pub/gnu_linux/centos/7.8.2003/isos/x86_64/CentOS-7-x86_64-Minimal-2003.iso) together with the appropriate sha256sum as a fix. Syntax changes in centos7-hyper-v.json required by packer version 1.6.0 - latest version are done using the output of the command "packer fix centos7-hyper-v.json".  

The installer is not tested!

### DIFF
--- a/packer/centos7-hyper-v.json
+++ b/packer/centos7-hyper-v.json
@@ -1,41 +1,30 @@
 {
-  "variables": {
-    "version": ""
-  },
-  "provisioners": [
-    {
-      "type": "shell",
-      "execute_command": "{{.Vars}} sudo -S -E bash '{{.Path}}'",
-      "script": "scripts/vagrant-setup.sh"
-    }
-  ],
   "builders": [
     {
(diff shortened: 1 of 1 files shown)


david-oc-miller (approved):

david-oc-miller:
@bojmil-armedia , @marjan4jc , to be honest I don't think developers should use the HyperV VM.  HyperV turns out to be much worse than VirtualBox, among other issues, HyperV killed my network performance (two orders of magnitude slower with HyperV enabled), and also, no shared clipboard support with Linux guests.  VirtualBox doesn't kill my network and has a shared clipboard.

PR #137, opened 2020-08-10
Added haproxy role that will replace httpd reverse proxy functionality

kikomanaskov333:
Here is the haproxy role that will replace httpd reverse proxy functionality. For now, this 'version' of role is only applicable on non-AWS environments.
Also we have to made changes into httpd role, where httpd will bind on different ports since haproxy use 443 and 80. For that reason here I'm using 488 port for FOIA and in my httpd (ssl.conf) i have: Listen 0.0.0.0:488 and 'VirtualHost _default_:488'. 
That's how the whole architecture works.


david-oc-miller:
@kikomanaskov333 I don't see any changes to the httpd role?  Is the new httpd port only configured in the facts file?  If so then the arkcase-ce-facts.yml and arkcase-dev-facts.yml probably should be updated?

kikomanaskov333:
@david-oc-miller , you can see in the new commits there are some changes in the httpd role

dzdimov (commented):
Do we keep the previous configuration data in httpd as commented or do we want to delete them?

bojmil-armedia:
We need to setup HaProxy entries in the microservices role. Also we can now proxy tcp protocols for elasticsearch and activemq, instead of opening those port to public.

vagrant/provisioning/roles/httpd/templates/httpd-alfresco-redirect.conf
```diff
@@ -277,24 +277,24 @@ SSLCertificateChainFile "{{ ssl_ca }}"
...
+#ProxyPass /alfresco https://{{ internal_host }}:7070/alfresco
+#ProxyPassReverse /alfresco https://{{ internal_host }}:7070/alfresco
 
-<LocationMatch "/alfresco">
-  <If '%{REMOTE_ADDR} != {{ arkcase_host_address | default('127.0.0.1') }}'>
-    RequestHeader unset X-Alfresco-Remote-User
-  </If>
-</LocationMatch>
+#<LocationMatch "/alfresco">
+#  <If '%{REMOTE_ADDR} != {{ arkcase_host_address | default('127.0.0.1') }}'>
```
david-oc-miller: Do we need this file any more?  If everything is commented out let's just delete the whole file.

vagrant/provisioning/roles/httpd/templates/httpd-ssl.conf
```diff
@@ -287,44 +287,44 @@ SSLCertificateChainFile "{{ ssl_ca }}"
 
 
 # Reverse proxy settings for Arkcase and components
-ProxyPass /solr https://{{ solr_host }}:8983/solr
-ProxyPassReverse /solr https://{{ solr_host }}:8983/solr
-<Location /solr>
-  Include conf.d/ldap-auth-restricted-solr
-</Location>
+#ProxyPass /solr https://{{ solr_host }}:8983/solr
+#ProxyPassReverse /solr https://{{ solr_host }}:8983/solr
```
david-oc-miller: Let's delete all these lines instead of commenting them out.

david-oc-miller:
@bojmil-armedia can you let @kikomanaskov333 know what configuration is needed for elastic and ActiveMQ (port, target)?  How about the Kafka REST proxy and whatnot?  Maybe we need a separate ticket for microservice haproxy config.

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -88,7 +88,6 @@
       
       frontend main
         mode http
-        bind *:80
```
david-oc-miller: do we need to bind to port 80 in order to auto-redirect to port 443 in case anybody uses http instead of https?

bojmil-armedia:
Yes, let's create another issue for microservices HaProxy changes and another one for the proxying TCP protocols for ActiveMQ and Kafka.

david-oc-miller:
> Yes, let's create another issue for microservices HaProxy changes and another one for the proxying TCP protocols for ActiveMQ and Kafka.

https://arkcase.atlassian.net/browse/DSO-38

https://arkcase.atlassian.net/browse/DSO-39

vagrant/provisioning/roles/haproxy/tasks/main.yml
david-oc-miller: @kikomanaskov333 , I still think we need to keep the bind to port 80, in order to send a redirect to port 443.

kikomanaskov333: @david-oc-miller , yes you're right, but then we need to change httpd default config (httpd.conf) not to bind to port 80.

david-oc-miller: @kikomanaskov333 , yes, let's change the httpd bindings, so we know that only haproxy manages the ports that are visible to outside.  The httpd ports should not be open externally but instead should only be reachable through haproxy.

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -0,0 +1,193 @@
...
+  become: yes
+  firewalld:
+    immediate: yes
+    permanent: yes
+    port: "{{ item }}"
+    state: enabled
+  loop:
+    - "80/tcp"
+    - "443/tcp"
+    - "488/tcp"
```
david-oc-miller: Do we really need to open port 488 itself?  I think the FOIA proxy config should be enough, so 488 doesn't have to be open in the firewall?

kikomanaskov333: Thank you @david-oc-miller , you are right. Actually we don't need to open port 488

david-oc-miller (approved):

PR #139, opened 2020-08-11
Uncommented the lines in the privacy portal role for the portal templ…

dzdimov:
I have uncomented the task in the privacy role for the configuration of the ldap-server.yaml file for the arkcase-ldap-configuration role

### DIFF
--- a/vagrant/provisioning/roles/privacy/tasks/main.yml
+++ b/vagrant/provisioning/roles/privacy/tasks/main.yml
@@ -93,20 +93,20 @@
     regexp: "acm-activemq"
     replace: "{{ activemq_host }}"
 
-#- name: configure LDAP
-#  include_role: 
-#    name: arkcase-ldap-config
-#  vars:
-#    old_config_id: privacyportal
-#    new_config_id: "privacyportal"
-#    autogenerate_user_id: true
-#    ldap_user_domain_bean_name: "privacyportal"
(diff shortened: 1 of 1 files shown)


bojmil-armedia (approved):

vagrant/provisioning/roles/privacy/tasks/main.yml
```diff
@@ -93,20 +93,20 @@
...
+    name: arkcase-ldap-config
+  vars:
+    old_config_id: privacyportal
+    new_config_id: "privacyportal"
+    autogenerate_user_id: true
+    ldap_user_domain_bean_name: "privacyportal"
+    ldap_authentication_service_class: "gov.privacy.service.SARLdapAuthenticationService"
+    portal_prefix: "portal."
+    ldap_template_name: "spring-config-privacyportal-ldap.xml"
+    group_base: "{{ ldap_portal_group_no_base | default(ldap_group_no_base) }}"
```
david-oc-miller, line 107: I suppose the privacy portal and the FOIA portal should have different OUs in case they are both being used in the same installation.

bojmil-armedia: We don't have that option right now. Privacy is a CaseFile extension, as is Foia. I don't think we can enable them both.

david-oc-miller: OK .. we'll worry about this when we actually have this requirement.

PR #141, opened 2020-08-17
Update Java version based on Snowbound version

bojmil-armedia:


dzdimov (approved):

vagrant/provisioning/roles/snowbound-app/tasks/main.yml
```diff
@@ -1,3 +1,12 @@
+- name: set JAVA_HOME
+  become: yes
+  become_user: "snowbound"
+  blockinfile:
+    path: "{{ root_folder }}/app/snowbound/bin/setenv.sh"
+    marker: "# {mark} ANSIBLE MANAGED BLOCK - JAVA_HOME"
+    block: |
```
david-oc-miller, line 7: The JAVA_HOME is already set in the Tomcat installation play that runs before this one.  Consider removing this change... I notice the snowbound role already has this line:

java_home: "{{ '/usr/lib/jvm/java-1.8.0-openjdk' if snowbound_vendor_version == '4.10' else '/usr/lib/jvm/java-11-openjdk' }}"

This is line 50 of vagrant/provisioning/roles/snowbound/tasks/main.yml.  

Why did we need this line?  It seems to be the same as the existing line?

bojmil-armedia: We have to have it here as well. If you change Snowbound version to 5.2, the upgrade script will update this version, without running the **snowbound** role. And also we can switch back to 4.10, if we are not satisfied or whatever...

david-oc-miller: @bojmil-armedia  why can't we run the snowbound role in the update?

bojmil-armedia: In theory we can run all roles, which would update everything. But it lasts for 1 hour. The shortened update roles run in 15 minutes. Perhaps, we can work on getting the time to re-run all roles and update everything shorter.

david-oc-miller: @bojmil-armedia how about a cron job that runs every day; pulls the installer code; and re-runs the installer if it pulls any changes? Or maybe runs every week?

dzdimov: In that case why don’t we pursue the idea of implementing terraform and destroy and create the vms?

david-oc-miller: @dzdimov how do you plan to backup and restore the application data?  In dev and test and almost all pre-production and production sites, the application VM also stores all the data.

dzdimov: I was thinking that we implement this approach only on dev in the beginning where I think we don’t need to backup the configuration.

david-oc-miller: Sometime we need to figure out stack upgrades in place, specially since apparently we need it for AWS Marketplace.

PR #143, opened 2020-08-19
Haproxy logging and logrotation

kikomanaskov333:


vagrant/provisioning/roles/haproxy/templates/rh-haproxy18-haproxy.logrotate.d
```diff
@@ -0,0 +1,10 @@
+{{ root_folder }}/log/haproxy/*.log {
+  copytruncate
+  daily
+  rotate 7
+  compress
+  missingok
+  size 10k
+  dateext
+  dateyesterday
```
david-oc-miller, line 9: The HAproxy logs get pretty large pretty quickly.  Please check on some of the dev servers after they get this change.  We may want to change the default configuration to only log warnings or errors.

This all looks good though.

david-oc-miller (approved):

PR #149, opened 2020-08-25
Allowed configuration of Solr max heap space

kikomanaskov333:


dzdimov (dismissed):

dzdimov:
Do we need a solr restart after this?

vagrant/provisioning/roles/solr/tasks/install-service.yml
```diff
@@ -55,6 +55,15 @@
...
       SOLR_SSL_WANT_CLIENT_AUTH=false
 
+- name: updated environment file - Solr max heap space
+  become: yes
+  become_user: solr
+  lineinfile:
+    path: "{{ root_folder }}/app/solr/bin/solr.in.sh"
+    backup: yes
+    regexp: '^.*SOLR_JAVA_MEM.*'
+    line: 'SOLR_JAVA_MEM="-Xms{{ solr_heap_mem | default("2048m") }} -Xmx{{ solr_heap_mem | default("2048m") }}"'
```
david-oc-miller: This block is not idempotent; the SOLR_JAVA_MEM line will be replaced every time the play is run.  We prefer idempotent plays, so if the memory is already set correctly, the line isn't written.

Also, I know in the ticket I said 2048m should be default, but on developer VMs 2G is not available.  Let's keep the same default as Solr (512m).  I'll update the ticket if you like.

david-oc-miller: And right, once the play is idempotent, then you can register a fact and restart solr if that fact is changed.

david-oc-miller (approved):

PR #154, opened 2020-09-02
Alfresco should have support of second LDAP configuration for the portal users

kikomanaskov333:
There are changes in the main.yml which they support second LDAP configuration only when we deploy FOIA

### DIFF
--- a/vagrant/provisioning/roles/alfresco/tasks/main.yml
+++ b/vagrant/provisioning/roles/alfresco/tasks/main.yml
@@ -374,6 +374,60 @@
       opencmis.server.override=true
       opencmis.server.value=https://{{ external_host }}/alfresco/api
 
+- name: change alfresco-global.properties to support second LDAP configuration
+  block:
+    - name: add ldap2 in AD/Samba section
+      become: yes
+      become_user: alfresco
+      replace:
+        path: "{{ root_folder }}/app/alfresco/shared/classes/alfresco-global.properties"
+        regexp: authentication\.chain=external1:external,ldap1:ldap-ad,alfrescoNtlm1:alfrescoNtlm
+        replace: authentication.chain=external1:external,ldap1:ldap-ad,ldap2:ldap-ad,alfrescoNtlm1:alfrescoNtlm
+    - name: remove LDAP host-specific settings section
+      become: yes
+      become_user: alfresco
+      replace:
+        path: "{{ root_folder }}/app/alfresco/shared/classes/alfresco-global.properties"
+        after: '#### LDAP host-specific settings'
+        before: '# remove unneeded Alfresco activity'
+        regexp: '.*'
+        replace: ''
+    - name: create ldap-ad folder structure
+      become: yes
+      become_user: alfresco
+      file:
+        path: "{{ item }}"
+        state: directory
+        recurse: yes
+      loop:
+        - "{{ root_folder }}/app/alfresco/shared/classes/alfresco/extension/subsystems/Authentication/ldap-ad/ldap1"
+        - "{{ root_folder }}/app/alfresco/shared/classes/alfresco/extension/subsystems/Authentication/ldap-ad/ldap2"
+    - name: create ldap1-ad.properties file and populate it 
+      become: yes
+      become_user: alfresco
+      copy:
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -374,6 +374,60 @@
       opencmis.server.override=true
       opencmis.server.value=https://{{ external_host }}/alfresco/api
 
+- name: change alfresco-global.properties to support second LDAP configuration
+  block:
+    - name: add ldap2 in AD/Samba section
```
david-oc-miller, line 384: This task is adding ldap1, but the task name mentions only ldap2.

david-oc-miller: Never mind, now I see what you're doing.

vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -374,6 +374,60 @@
...
+      become: yes
+      become_user: alfresco
+      replace:
+        path: "{{ root_folder }}/app/alfresco/shared/classes/alfresco-global.properties"
+        regexp: authentication\.chain=external1:external,ldap1:ldap-ad,alfrescoNtlm1:alfrescoNtlm
+        replace: authentication.chain=external1:external,ldap1:ldap-ad,ldap2:ldap-ad,alfrescoNtlm1:alfrescoNtlm
+    - name: remove LDAP host-specific settings section
+      become: yes
+      become_user: alfresco
+      replace:
```
david-oc-miller: Why not just remove this part of the earlier alfresco-global.properties template?  Why should the alfresco role first add LDAP host-specific settings, and then remove them?

vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -374,6 +374,60 @@
...
+      become: yes
+      become_user: alfresco
+      file:
+        path: "{{ item }}"
+        state: directory
+        recurse: yes
+      loop:
+        - "{{ root_folder }}/app/alfresco/shared/classes/alfresco/extension/subsystems/Authentication/ldap-ad/ldap1"
+        - "{{ root_folder }}/app/alfresco/shared/classes/alfresco/extension/subsystems/Authentication/ldap-ad/ldap2"
+    - name: create ldap1-ad.properties file and populate it
```
david-oc-miller: I still think we should always create the ldap1 folder with these properties, then only create ldap2 with its properties if we are deploying the foia portal.

david-oc-miller: Also I think you can add a line to set the sync schedule, let's sync at least every hour.

bojmil-armedia: I agree with creating ldap1 folder always.
The sync should be set on 5 minutes. Otherwise, the user cannot do anything until it is synced to Alfresco. We didn't find a way to force Alfresco sync when a new user logs in to Arkcase (for Alfresco CE version).

david-oc-miller: @kikomanaskov333  please add following to alfresco-global.properties:

nchronization.syncWhenMissingPeopleLogIn=true
synchronization.import.cron=0 0/5 * * * ? *

vagrant/provisioning/roles/alfresco/tasks/main.yml
kikomanaskov333: @david-oc-miller, by default in case when we're using non foia installation the LDAP host-specific should be defined in alfresco-global.properties. In case of foia installation i have to remove that block (LDAP host-specific). You can see that this part will be running only when the condition 'foia_portal_context is defined' is satisfied, otherwise it will be skipped.

kikomanaskov333: From the comments bellow, if we always create the ldap1 folder than i will made a changes according to that

david-oc-miller: @kikomanaskov333, exactly; we always create ldap1 and put the ArkCase LDAP info there; so we never need the LDAP host-specific info in the alfresco-global.properties.  In case foia_portal_context is defined, then we also create an ldap2 folder, but we always create ldap1.

david-oc-miller (approved):

PR #156, opened 2020-09-03
Dso 64 disable admin in playbook

fabricetriboix:


kikomanaskov333 (approved):

vagrant/provisioning/arkcase-ce-foia.yml
```diff
@@ -54,4 +54,5 @@
       tags: [core]
     - role: start-arkcase
       tags: [core, arkcase]
-    
+    - role: disable-admin
```
david-oc-miller, line 57: Who is going to re-enable arkcase-admin in non-AWS FOIA installations?

fabricetriboix: @david-oc-miller That will be the startup script: https://github.com/ArkCase/DevOps/tree/develop/ArkCaseAMI-Marketplace

kikomanaskov333: @david-oc-miller you are right. @fabricetriboix we didn't take into consideration that in non-AWS FOIA installations the script is not using. We have to made additional play that will be used only for FOIA AWS installations. The arkcase-ce-foia.yml is using everywhere

PR #163, opened 2020-09-15
The LDAP should support additional OU for FOIA

kikomanaskov333:
Changes in ldap main.yml in order the ldap to support addition OU for FOIA.
Also the facts are defined in the arkcase-dev-facts.yml

### DIFF
--- a/vagrant/provisioning/roles/ldap/tasks/main.yml
+++ b/vagrant/provisioning/roles/ldap/tasks/main.yml
@@ -57,7 +57,24 @@
   loop:
     - "{{ ldap_group_base }}"
     - "{{ ldap_user_base }}"
-    
+
+- name: add portal organizational units
+  ldap_entry:
+    server_uri: "{{ ldap_url }}"
+    validate_certs: no
+    bind_dn: "{{ ldap_bind_user }}"
+    bind_pw: "{{ ldap_bind_password }}"
+    dn: "{{ item }}"
+    objectClass:
+      - organizationalUnit
+      - top
+    attributes:
+      description: "{{ item }}"
+  loop:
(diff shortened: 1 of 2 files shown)


dzdimov (approved):

vagrant/provisioning/roles/ldap/tasks/main.yml
```diff
@@ -73,7 +90,42 @@
...
+- name: remove portal group from ArkCase OU if exists
+  ldap_entry:
+    server_uri: "{{ ldap_url }}"
+    validate_certs: no
+    bind_dn: "{{ ldap_bind_user }}"
+    bind_pw: "{{ ldap_bind_password }}"
+    dn: "CN={{ item.name }},{{ ldap_group_no_base }},{{ ldap_base }}"
+    state: absent
+    objectClass:
+      - group
```
david-oc-miller, line 103: What will happen to any users who might be in this group that you are removing?  Shall we first add the new portal group, and ensure to add any members from this old portal group, and then remove it?

vagrant/provisioning/roles/ldap/tasks/main.yml
```diff
@@ -57,7 +57,24 @@
...
+    bind_dn: "{{ ldap_bind_user }}"
+    bind_pw: "{{ ldap_bind_password }}"
+    dn: "{{ item }}"
+    objectClass:
+      - organizationalUnit
+      - top
+    attributes:
+      description: "{{ item }}"
+  loop:
+    - "{{ ldap_portal_group_no_base }},{{ ldap_base }}"
```
david-oc-miller, line 74: This block will work, but you could also just add these two items to the existing block that already adds OUs. This block isn't actually necessary.

PR #166, opened 2020-09-17
Changes in haproxy.conf in case when we will not deploy FOIA

kikomanaskov333:
All the configurations for FOIA presented in the frontend part are removed in case when we will deploy only Core

### DIFF
--- a/vagrant/provisioning/roles/haproxy/tasks/main.yml
+++ b/vagrant/provisioning/roles/haproxy/tasks/main.yml
@@ -140,7 +140,6 @@
         acl acl_alfresco path_beg /alfresco
         acl acl_share path_beg /share
         acl acl_snowbound path_beg /VirtualViewerJavaHTML5
-        acl acl_foia path_beg /foia
         acl acl_arkcase path_beg /arkcase
         
         # ensure internal sites only visible internally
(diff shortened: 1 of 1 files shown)


bojmil-armedia (approved):

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -194,7 +192,25 @@
         server arkcase0 {{ arkcase_host }}:8843 check fall 3 rise 2 check ssl verify required ca-file {{ ssl_ca }}
   
   register: arkcase_config
-  
+
+- name: FOIA fronted part
```
david-oc-miller, line 196: minor typo... fronted -> frontend

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -194,7 +192,25 @@
...
+  become: yes
+  lineinfile:
+    backup: yes      
+    path: "/etc/opt/rh/rh-haproxy18/haproxy/haproxy.cfg"
```
david-oc-miller, line 200: Please verify what happens if you run this play multiple times. I'm pretty sure it will add these lines every time you run the play (run the play 5 times and you will have 5 acl_foia acl lines, and 5 use_backend be_foia lines).

If I'm right you can make this play idempotent by first detecting if the lines already exist.

PR #167, opened 2020-09-21
Changed the facts file for the local developers VM's to enable the el…

dzdimov:
…asticsearch and confluent servies

By default the confluent services were not started on the developers VM's.
By doing this I have enabled them and they should have no problems anymore.

Furthermore this might cause issues with the developers who won't be working on ADK since their current laptops might not support all the services running due to limited RAM resources. 

If someone can suggest another approach we can make the changes accordingly.

### DIFF
--- a/vagrant/provisioning/arkcase-dev-facts.yml
+++ b/vagrant/provisioning/arkcase-dev-facts.yml
@@ -500,5 +500,5 @@ confluent_scala_version: 2.12
 
 zipkin_version: 2.21.4
 # Enable these for Arkcase.NEXT development
-enable_elasticsearch: "no"
-enable_kafka: "no"
+enable_elasticsearch: "yes"
+enable_kafka: "yes"


vagrant/provisioning/arkcase-dev-facts.yml
```diff
@@ -500,5 +500,5 @@ confluent_scala_version: 2.12
 
 zipkin_version: 2.21.4
 # Enable these for Arkcase.NEXT development
-enable_elasticsearch: "no"
-enable_kafka: "no"
+enable_elasticsearch: "yes"
```
david-oc-miller, line 503: I believe there's a separate facts file and corresponding playbook something like "arkcase-microservices-facts.yml" and "arkcase-microservices.yml". Make these changes there only (not in this file which is used by core/FOIA developers) and all should be well, I believe.

dzdimov: I see what the issue is.

The developers were following these instructions: https://gitlab.armedia.com/arkcase/ArkCase/wikis/ArkCaseCE-on-Windows
instead of https://gitlab.armedia.com/arkcase/ArkCase/wikis/ArkCase-on-Windows in which we have a part in which we say 
"
1.3. *(Optional) If developing ArkCase.Next microservices, edit the Vagrantfile and change the line:  
`ansible-playbook -e @arkcase-dev-facts.yml -e "sftp_arkcase_password=*****" arkcase-dev.yml`  
to  
`ansible-playbook -e @arkcase-dev-microservices-facts.yml -e "sftp_arkcase_password=*****" arkcase-dev-microservices.yml`"

hence why the developers are reporting that the confluent services aren't starting.

I will close this MR as this will not be needed since in the arkcase-microservices-facts.yml the fact for the confluent services are already as they should be

dzdimov:
I will close this MR as this will not be needed since in the arkcase-microservices-facts.yml the fact for the confluent services are already as they should be

vagrant/provisioning/arkcase-dev-facts.yml
david-oc-miller: Can you update the ArkCaseCE-on-windows wiki to include the same comment in terms of using arkcase-dev-microservices-facts.yml?

dzdimov: Wiki is now updated

PR #168, opened 2020-09-22
Changed the Solr service to be a native systemd service

dzdimov:
We previously had the SOLR service file masked as a systemd file but invoking init.d service which in turn did not report his true state to systemd hence after crash the service would not restart by itself.

We have changed the ExecStart and ExecStop to point to /solr/bin/solr and not to the init.d file.
By doing this change the solr service will now be a native systemd service.


kikomanaskov333 (dismissed):

bojmil-armedia (dismissed):

vagrant/provisioning/roles/solr/templates/solr.service
```diff
@@ -1,17 +1,12 @@
...
-RemainAfterExit=true
-ExecStart={{ root_folder }}/app/solr/bin/init.d/solr start
-ExecStop={{ root_folder }}/app/solr/bin/init.d/solr stop
-# Starting from CentOS 7.7 we can't specify the PIDfile for 
-# solr anymore; but systemd can find it from the process 
-# table anyway.
-# PIDFile={{ root_folder }}/app/solr/bin/solr-8983.pid
 User=root
 Group=root
+PIDFile=/opt/app/arkcase/app/solr/bin/solr-8983.pid
```
fabricetriboix: I guess we shouldn't use a hardcoded path here, but `{{ root_folder }}` instead?

vagrant/provisioning/roles/solr/templates/solr.service
```diff
@@ -1,17 +1,12 @@
...
 [Service]
 Type=forking
-RemainAfterExit=true
-ExecStart={{ root_folder }}/app/solr/bin/init.d/solr start
-ExecStop={{ root_folder }}/app/solr/bin/init.d/solr stop
-# Starting from CentOS 7.7 we can't specify the PIDfile for 
-# solr anymore; but systemd can find it from the process 
-# table anyway.
-# PIDFile={{ root_folder }}/app/solr/bin/solr-8983.pid
 User=root
```
fabricetriboix: I think the user should `solr`, not `root`

fabricetriboix (changes requested):

vagrant/provisioning/roles/solr/templates/solr.service
david-oc-miller: definitely need {{ root_folder }} here.

vagrant/provisioning/roles/solr/templates/solr.service
david-oc-miller: Somehow solr requires to start as root user, then it forks a child process as the user indicated in the solr.in.sh file.

dzdimov: Changed the user to solr as well

fabricetriboix: @david-oc-miller When I tested that with @dzdimov , I actually had to set the user to `solr` in the unit file. @dzdimov Have you tested it with the `solr` user?

david-oc-miller: Well, maybe the requirement to start as root was imposed by the /etc/init.d-style script.  Now that we call solr directly you probably do have to run as solr user.  This revised service unit is a huge improvement over the old one!

kikomanaskov333 (approved):

PR #169, opened 2020-09-24
Copy the arkcase-portal-server.yaml from backup

dzdimov:
Here I am taking the same approach as restoring the branding files. I get the latest .arkcase backup and get the arkcase-portal-server.yaml from that location and copy it to the new .arkcase configuration folder, thus retaining the old foia portal configuration. 
Also this task will only execute if the installation is foia based and a flag "ignore_errors: true" is set in case this is a first time installation in which case we won't have a previous backup from which the ansible play would take the file.

### DIFF
--- a/vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
@@ -31,3 +31,13 @@
     - "custom-runtime.css"
   when: latest_config_backup_folder is defined
   ignore_errors: true
+
+ - name: Restore arkcase portal configuration
+   become: yes
+   become_user: arkcase
+   command: cp -f {{ latest_config_backup_folder.path }}/acm/acm-config-server-repo/{{ item }} {{ root_folder }}/data/arkcase-home/.arkcase/acm/acm-config-server-repo/{{ item }}
+   loop:
+     - "arkcase-portal-server.yaml"
+     - "arkcase-portal-server.yaml"
+   when: foia_portal_context is defined
+   ignore_errors: true
--- a/vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
@@ -38,6 +38,5 @@
    command: cp -f {{ latest_config_backup_folder.path }}/acm/acm-config-server-repo/{{ item }} {{ root_folder }}/data/arkcase-home/.arkcase/acm/acm-config-server-repo/{{ item }}
    loop:
      - "arkcase-portal-server.yaml"
-     - "arkcase-portal-server.yaml"
    when: foia_portal_context is defined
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
```diff
@@ -31,3 +31,13 @@
...
   when: latest_config_backup_folder is defined
   ignore_errors: true
+
+ - name: Restore arkcase portal configuration
+   become: yes
+   become_user: arkcase
+   command: cp -f {{ latest_config_backup_folder.path }}/acm/acm-config-server-repo/{{ item }} {{ root_folder }}/data/arkcase-home/.arkcase/acm/acm-config-server-repo/{{ item }}
+   loop:
+     - "arkcase-portal-server.yaml"
+     - "arkcase-portal-server.yaml"
```
fabricetriboix: Why is there twice the same file?

fabricetriboix (dismissed):

vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
```diff
@@ -31,3 +31,13 @@
     - "custom-runtime.css"
   when: latest_config_backup_folder is defined
   ignore_errors: true
+
+ - name: Restore arkcase portal configuration
+   become: yes
+   become_user: arkcase
+   command: cp -f {{ latest_config_backup_folder.path }}/acm/acm-config-server-repo/{{ item }} {{ root_folder }}/data/arkcase-home/.arkcase/acm/acm-config-server-repo/{{ item }}
```
bojmil-armedia, line 38: We should not get the latest backup completely. The **arkcase-portal-server.yaml** template might have new/changed properties.  
We should only get the **portal.authenticatedMode** property from the backup file and use it in the template.

vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
dzdimov: Thanks for noticing. It was somehow copied twice

kikomanaskov333 (approved):

bojmil-armedia (commented):
We should not get the latest backup completely. The arkcase-portal-server.yaml template might have new/changed properties.
We should only get the portal.authenticatedMode property from the backup file and use it in the template.

vagrant/provisioning/roles/arkcase-post-upgrade/tasks/main.yml
david-oc-miller: I guess from a long term perspective we need to add "better upgrade support" to our technical roadmap.  I hacked together the current approach to support our dev and test environments and for the most part this approach seems to have resolved most of the "test-environment-only" issues.  Let's consider a transactional approach like our Liquibase change sets, where we build up our configurations line-item by line-item, or anyway come up with something better than this.

PR #172, opened 2020-10-07
Admin Configuration for setting Authenticated or UN-Authenticated mode

bojmil-armedia:


NCuculova (approved):
Looks good.

vagrant/provisioning/roles/foia/templates/arkcase-portal-server-2020.16.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+
+portal.externalConfiguration: true
+portal.host: {{ external_host }}
+portal.port:
+portal.uiApplicationUrl: "https://{{ external_host }}/foia"
+portal.uiApplicationHost: "https://{{ external_host }}"
+portal.id: {{ 9999999999999999999999 | random | to_uuid }}
+
+portal.authenticatedMode: false
```
david-oc-miller, line 17: I can't help but notice that `portal.authenticatedMode` is set to `false` and I don't see anything that would change it to `true`.  Am I missing anything?

david-oc-miller: Or is this configuration somehow changed to the ArkCase admin page can manage `portal.authenticatedMode`?

bojmil-armedia: Yes. It can be changed now from admin page. Although, I would add this as a variable for different servers. And there is a DSO ticket for that.

PR #176, opened 2020-10-14
Improving ansible syntax for HaProxy

kikomanaskov333:
The following changes have been made:

- Blokinfile was missing under each block. Now it is added everywhere
- 'register' attribute is removed since it is not a valid attribute for a Block.

### DIFF
--- a/vagrant/provisioning/roles/haproxy/tasks/main.yml
+++ b/vagrant/provisioning/roles/haproxy/tasks/main.yml
@@ -233,190 +233,198 @@
   block:
     - name: Populate haproxy.conf with fronted part 
       become: yes
-      marker: "# {mark} ANSIBLE MANAGED BLOCK ACL CONFLUENT PLATFORM CONTEXT"
-      insertafter: "#arkcase.next services"
-      path: "/etc/opt/rh/rh-haproxy18/haproxy/haproxy.cfg"
-      block: |
-        acl acl_schema-registry path_beg /schema-registry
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -438,14 +446,14 @@
...
     daemon_reload: yes
-  when: haproxy_service_name is changed
-
-- name: enable haproxy to connect to any port
-  become: yes
-  seboolean:
-    name: haproxy_connect_any
-    state: yes
-    persistent: yes
+  when: haproxy_serivce_name is changed
```
david-oc-miller: Here you introduced a typo; it should remain haproxy_service_name versus haproxy_serivce_name

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -438,14 +446,14 @@
...
+
+#- name: enable haproxy to connect to any port
```
david-oc-miller: why is this commented out?  Haproxy won't connect to local ports without this option.

kikomanaskov333: Ohh, I forgot to uncomment. It was remain commented out since on my environment the selinux is disabled and that part failed when your selinux is disabled. I'll fix it

bojmil-armedia (approved):

PR #182, opened 2020-10-21
Removed HTTPD proxy settings for microservices

kikomanaskov333:


vagrant/provisioning/roles/microservices/tasks/main.yaml
```diff
@@ -9,14 +9,6 @@
   loop:
     - "9092/tcp"   # Kafka
 
-- name: Add HTTPD proxy settings for microservices
-  become: yes
```
david-oc-miller, line 13: I guess this is because microservices are routed via haproxy exclusively now, and not through httpd at all?

david-oc-miller (approved):

vagrant/provisioning/roles/microservices/tasks/main.yaml
kikomanaskov333: Yes, exactly @david-oc-miller

PR #188, opened 2020-10-29
DSO-193 Zabbix server role

david-oc-miller:


david-oc-miller:
We are not proceeding with Zabbix.

PR #189, opened 2020-10-29
DSO-194 SAML and FOIA improvements

david-oc-miller:


vagrant/provisioning/roles/foia/templates/application.properties
```diff
@@ -13,15 +13,15 @@ portal.host={{ external_host }}
 portal.port=
 portal.uiApplicationUrl=https://{{ external_host }}/foia
 portal.uiApplicationHost=https://{{ external_host }}
-portal.id={{ 9999999999999999999999 | random | to_uuid }}
+portal.id={{ foia_portal_id | default(9999999999999999999999 | random | to_uuid) }}
```
david-oc-miller, line 16: after accepting this change (assuming we go ahead) the facts file for existing portal sites should be updated to include the `foia_portal_id`, `portal_authenticated_mode` and `foia_portal_authentication_type` facts.

vagrant/provisioning/roles/foia/templates/application.properties
```diff
@@ -41,15 +41,15 @@ portal.truststoreFilename=arkcase.ts
...
+portal.authenticationType={{ foia_portal_authentication_type | default('basic') }}
 # basic
 portal.authenticationBasicUser={{ arkcase_admin_user }}@{{ ldap_user_domain }}
 portal.authenticationBasicPassword={{ arkcase_admin_password }}
 # external
 portal.authenticationExternalUser={{ arkcase_admin_user }}@{{ ldap_user_domain }}
 
 # ArkCase url
-portal.arkcaseUrl=https://{{ internal_host }}/arkcase
+portal.arkcaseUrl=https://{{ internal_host }}:8843/arkcase
```
bojmil-armedia, line 52: This one should work through the proxy without a port.

bojmil-armedia (approved):

vagrant/provisioning/roles/foia/templates/application.properties
david-oc-miller: In multi-node environments the proxy isn't local.  I could have tried using external_host versus internal_host, and maybe I'll see if that will work next.  This way seemed easier, and avoids a round trip to the proxy.  Of course if the portal ever ends up on a different host than ArkCase we'll need to use external_host anyway.

PR #192, opened 2020-11-02
Use arkcase-ee variable

bojmil-armedia:
**arkcase-ee: true** fact should be added to all templates.


bojmil-armedia:
Created ticket https://arkcase.atlassian.net/browse/DSO-203 to arkcase-ee: true fact should be added to all templates.

david-oc-miller (approved):

PR #197, opened 2020-11-04
Update main.yml

dzdimov:
Updated absolute path with relative


fabricetriboix (approved):

david-oc-miller:
Guys, as a general rule, let's use specific merge request titles: "fix path to pentaho file" is more clear than "update main.yml" when somebody is reviewing the list of merge requests.

PR #203, opened 2020-11-06
Consolidate all the playbooks

kikomanaskov333:
Here are two playbooks which contain all the roles inside. 
In the arkcase-all.yml you can see there are also presented the roles that have been doing upgrade.


david-oc-miller:
I like the approach overall.  I don't think we need separate CE and EE playbooks after all.

bojmil-armedia:
Looks good. Some points:
1. Delete all other templates
2. Why a separate AWS play?
3. The dev playbook should be named as it is now. No need to change our instructions.
4. The dev-microservices playbook, should be existing or create a new microservices playbook and we'll just ask microservices developer to run this role additionally.
5. In comments add the combinations of tags we are currently using.

vagrant/provisioning/arkcase-all-AWS.yml
```diff
@@ -0,0 +1,74 @@
...
+      tags: [foia, foia-analytical-reports]
+    - role: tesseract
+      tags: [core, arkcase, tesseract]
+    - role: firewall
+      when: enable_firewall is undefined or enable_firewall
+      tags: [core]
+    - role: saml-config
+      tags: [upgrade-saml]
+    - role: start-arkcase
+      tags: [core, arkcase, upgrade]
```
dzdimov: @kikomanaskov333 please change the order of the roles so that start-arkcase is before the foia-analytical-report role since the role will fail if ArkCase is not running

vagrant/provisioning/arkcase-all.yml
```diff
@@ -0,0 +1,72 @@
...
+    - role: foia-analytical-reports
+      tags: [foia, foia-analytical-reports]
+    - role: tesseract
+      tags: [core, arkcase, tesseract]
+    - role: firewall
+      when: enable_firewall is undefined or enable_firewall
+      tags: [core]
+    - role: saml-config
+      tags: [upgrade-saml]
+    - role: start-arkcase
```
dzdimov, line 85: You should also do the same here

kikomanaskov333:
Please take a look over the changes I did. You can find also the txt file with all the instruction of how to use the two playbooks (arkcase-all and arkcase-dev-all).
Here is a short description:
1. In the arkcase-all you can find all the roles for arkcase-ce, arkcase-ee , upgrades, and for the AWS Marketplace deployments. All you need is to invoke arkcase-all.yml with specific tags for the specific roles. The tags for each role you can find in the .txt file attached here.
2. The some approach I used for arkcase-dev-all.yml

All suggestions are warmly welcome :)

bojmil-armedia:
Looks good.
Not sure when to delete the old playbooks. Should keep them for backward compatibility for some time, until we migrate all templates? We can open a task to migrate templates to the new playbook.

bojmil-armedia (approved):

dzdimov (approved):

PR #208, opened 2020-11-17
Made changes to the arkcase.service

dzdimov:
Modified the arkcase.service with variables depending on the ldap_type: file
Added ldap_type fact
Modified the arkcase-app main yaml with the set_fact task which will be used to populate the arkcase_required_services


bojmil-armedia (approved):

david-oc-miller:
@dzdimov let's keep trying to improve the merge request names.  Consider `ArkCase: only depend on Samba if Samba is actually being used` compared to `Made changes to the arkcase.service`.

PR #209, opened 2020-11-18
Arkcase interactive reports deployment

bojmil-armedia:


vagrant/provisioning/arkcase-app-upgrade.yml
```diff
@@ -11,6 +11,8 @@
       tags: [core]
     - role: foia
       tags: [foia]
+    - role: arkcase-interactive-reports
+      tags: [core]
```
bojmil-armedia, line 15: @kikomanaskov333 please add this role in the new playbook

dzdimov (approved):

PR #227, opened 2020-12-03
Added new portal related facts into arkcase-dev-microservices-facts.yml

kikomanaskov333:


dzdimov (approved):

bojmil-armedia:
This is really not needed for developers, since the portal is run from code and that role isn't run. Anyways, it's ok to have them here as documentation.

PR #229, opened 2020-12-03
Changed the order of roles

dzdimov:
We need ArkCase to be running before deploying the reports


bojmil-armedia (approved):

bojmil-armedia:
@dzdimov please add this role to all playbooks that install the foia-analytical-reports

PR #231, opened 2020-12-07
Added support for dual ldap (separate ldap for arkcase and portal)

dzdimov:
This is the change that is done in order to support 2 ldaps-one for ArkCase(AD) and the other for the FOIA portal(samba)

We need to add these facts:
ldap_portal_base:
ldap_bind_portal_user: 
ldap_portal_bind_password: 
ldap_portal_url: 
ldap_portal_user_domain:

### DIFF
--- a/vagrant/provisioning/roles/ldap-portal/tasks/main.yml
+++ b/vagrant/provisioning/roles/ldap-portal/tasks/main.yml
@@ -1,0 +1,78 @@
+- name: LDAP required packages
+  become: yes
+  yum:
+    state: installed
+    name: 
+      - openldap-clients
+      - python-ldap
+
+- name: install pyOpenSSL
+  become: yes
+  pip:
+    name: pyOpenSSL
+
+- name: list certs in the key store
+  become: yes
+  command: keytool -v -list -keystore "{{ java_trust_store }}" -storepass "{{ java_trust_store_pass }}"
+  register: cert_list
+  changed_when: false
+
+- name: import the ldap cert to the ArkCase trust store
+  become: yes
+  java_cert:
+    cert_alias: "arkcase_ldap_cert"
+    cert_url: "{{ ldap_portal_host }}"
+    cert_port: "{{ ldap_portal_port|int }}"
+    keystore_path: "{{ java_trust_store }}"
+    keystore_pass: "{{ java_trust_store_pass }}"
+    state: present
+  when: "'arkcase_ldap_cert' not in cert_list.stdout"
+
+- name: add arkcase ancestors
+  ldap_entry:
+    server_uri: "{{ ldap_portal_url }}"
+    validate_certs: no
+    bind_dn: "{{ ldap_portal_bind_user }}"
+    bind_pw: "{{ ldap_portal_bind_password }}"
+    dn: "{{ item }}"
+    objectClass:
+      - organizationalUnit
+      - top
+    attributes:
+      description: "{{ item }}"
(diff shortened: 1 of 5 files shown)


vagrant/provisioning/roles/ldap-portal/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+- name: import the ldap cert to the ArkCase trust store
+  become: yes
+  java_cert:
+    cert_alias: "arkcase_ldap_cert"
+    cert_url: "{{ ldap_portal_host }}"
+    cert_port: "{{ ldap_portal_port|int }}"
+    keystore_path: "{{ java_trust_store }}"
+    keystore_pass: "{{ java_trust_store_pass }}"
+    state: present
+  when: "'arkcase_ldap_cert' not in cert_list.stdout"
```
david-oc-miller: Use a different alias here, we can't use `arkcase_ldap_cert` as alias for both the platform LDAP and the portal LDAP.

vagrant/provisioning/roles/ldap-portal/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
+- name: LDAP required packages
```
david-oc-miller, line 1: Somehow it would be better to avoid duplicating code by extracting the common blocks to a task that can be reused by both the platform LDAP and portal LDAP roles.  We can add this ticket to be done later if you can't do it in time for this installation.

vagrant/provisioning/roles/ldap-portal/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+    validate_certs: no
+    bind_dn: "{{ ldap_portal_bind_user }}"
+    bind_pw: "{{ ldap_portal_bind_password }}"
+    dn: "{{ item }}"
+    objectClass:
+      - organizationalUnit
+      - top
+    attributes:
+      description: "{{ item }}"
+  loop: "{{ ldap_ancestor_ous }}"
```
david-oc-miller, line 43: since we are splitting the platform and portal roles, we can also split the OU structure, since the platform LDAP and portal LDAP are not required to have the same OU structure.  Maybe `ldap_ancestor_ous` can be only the platform OUs, and `portal_ancestor_ous` could be the portal OU structure.

bojmil-armedia: We've talked about this and decided that this is only for our internal Samba. Generally, we (or the client) will create the OUs manually on the LDAP server, add groups and existing users manually to these OUs.

vagrant/provisioning/roles/ldap-portal/tasks/main.yml
```diff
@@ -20,13 +20,13 @@
...
   java_cert:
-    cert_alias: "arkcase_ldap_cert"
+    cert_alias: "arkcase_ldap_portal_cert"
     cert_url: "{{ ldap_portal_host }}"
     cert_port: "{{ ldap_portal_port|int }}"
     keystore_path: "{{ java_trust_store }}"
     keystore_pass: "{{ java_trust_store_pass }}"
     state: present
-  when: "'arkcase_ldap_cert' not in cert_list.stdout"
+  when: "'arkcase_ldap_portal_cert' not in cert_list.stdout"
```
david-oc-miller, line 29: looks good

kikomanaskov333 (approved):

PR #236, opened 2020-12-10
New approach for default values of some variables

kikomanaskov333:


dzdimov (dismissed):

vagrant/provisioning/roles/foia/tasks/main.yml
```diff
@@ -123,6 +123,31 @@
...
+  set_fact:
+    ldap_portal_bind_user: "{{ ldap_bind_user }}"
+  when: ldap_portal_bind_user is not defined
+
+- name: Set encrypted_ldap_portal_bind_password
+  set_fact:
+    encrypted_ldap_portal_bind_password: "{{ encrypted_ldap_bind_password }}"
+  when: encrypted_ldap_portal_bind_password is not defined
+
+- name: Set ldap_portal_url
```
david-oc-miller, line 141: what was wrong with the old way?

david-oc-miller (dismissed):

vagrant/provisioning/roles/foia/tasks/main.yml
kikomanaskov333: @david-oc-miller, somehow when the first var is equal by name with the var defined in the default section (for ex:  ldap_url: "{{ ldap_portal_url | default(ldap_url) }}" -> ldap_url =  ldap_url), the ansible throws an exception and for that reason we decided to go with defining facts as a new tasks for all such cases.

dzdimov (approved):

PR #240, opened 2020-12-14
Dso 269

dzdimov:
As this will be used only for the DOJ environments for the foia.gov integration the default value should be empty


david-oc-miller:
The MR changes look good, but please provide a more descriptive title.

bojmil-armedia (approved):

PR #242, opened 2020-12-16
Fix proxy pass to zipkin and elasticsearch

NCuculova:


vagrant/provisioning/roles/microservices/templates/zipkin.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Zipkin Server
-After=syslog.target
+After=syslog.target,elasticsearch
```
david-oc-miller: The After services should be space-separated vs comma-separated ... I believe I got this wrong at the very start.  But the docs are clear about the space separation.

NCuculova: Thanks, I'll make the fix. I didn't have a chance to test it, since elasticsearch always fails to start on "vagrant up".

david-oc-miller: hmm why is it always failing?  Let me check on my VM.

david-oc-miller (approved):

vagrant/provisioning/roles/microservices/templates/zipkin.service
david-oc-miller: If ArkCase is installed using arkcase-dev-microsevices-facts.yml than elastic should be enabled and it should start on vagrant up. I installed with arkcase-dev-facts.yml and the elastic service wasn't enabled, hence it didn't start.

PR #247, opened 2021-01-18
Removed the foia-analytical-reports

dzdimov:
I am not sure that we need the foia analytical reports for the ArkCase EE Core version so I am removing them from the playbook for now.


kikomanaskov333 (approved):
I think we should start using one or two generic playbooks that contain all the roles inside and just invoke them with specific tags

PR #250, opened 2021-01-29
Support for python3

kikomanaskov333:
You can see in my MR all that changes that have been made in order the installer to support python3


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -60,14 +60,15 @@
 # might have to upgrade PIP itself before installing openpyxl
 - name: update PIP if necessary
   become: yes
-  pip:
+  pip:
```
fabricetriboix, line 63: Why not add `executable: pip3` here as well? I guess the `pip` module by itself will upgrade pip2, not pip3...

dzdimov (approved):

PR #255, opened 2021-02-15
Modify the installer to use cryptography module instead of pyOpenSSL

jovestefanovski:
Modifying the installer to use python cryptography module instead of pyOpenSSL because the installer crashed every time with the newer versions of ansible and as a recommendation we switch to cryptography module.

### DIFF
--- a/vagrant/provisioning/roles/pki/tasks/main.yml
+++ b/vagrant/provisioning/roles/pki/tasks/main.yml
@@ -1,7 +1,7 @@
-- name: install pyOpenSSL
+- name: install cryptography
   become: yes
   pip:
-    name: pyOpenSSL
+    name: cryptography
 
 - name: create pki folder structure
   become: yes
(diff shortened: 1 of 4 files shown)


fabricetriboix (approved):
Looks good to me

dzdimov (approved):

vagrant/provisioning/roles/pki/tasks/main.yml
```diff
@@ -133,7 +133,7 @@
     type: RSA
     size: 2048
     mode: 0644
-    cipher: "AES-256-ECB"
+    cipher: "auto"
```
bojmil-armedia, line 136: What's with auto vs AES-256-ECB? The specific cipher is better, I believe. What if auto changes the cipher? Perhaps, a better/upgraded cipher would be used, but can this lead to issues like using a cipher that is not supported by some tool using the key?

dzdimov: Since we are changing pyOpenSSL with Cryptography the only option we can use here is auto since the installer will fail

Note from Ansible module overview page: https://docs.ansible.com/ansible/latest/collections/community/crypto/openssl_privatekey_module.html
**When using the cryptography backend, use auto.**

PR #257, opened 2021-02-17
Include the AWS credentials for transcribe feature

jovestefanovski:
Modify the installer to include AWS credentials if they are present in the facts file.


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -798,7 +798,31 @@
       application.rolesToGroups:
         {{ roles_to_groups_yaml | to_nice_yaml | indent(2) }}
   when: arkcase_version == "" or arkcase_version is version('3.3.3', '>=')
-    
+
+- name: Make sure AWS credentials are insterted in the arkcase-server.yaml file if they are defined
```
david-oc-miller: typo: should be `inserted` instead of `insterted`

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -798,7 +798,31 @@
...
+  become: yes
+  become_user: arkcase
+  lineinfile:
+    path: /opt/arkcase/data/arkcase-home/.arkcase/acm/acm-config-server-repo/arkcase-server.yaml
+    state: present
+    backup: True
+    backrefs: True
+    regexp: '^awsAccessKeyId'
```
david-oc-miller: Typically the fact names are formatted as `aws_access_key_id` and they don't end in `fact`.

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -798,7 +798,31 @@
...
+    line: 'awsAccessKeyId: {{ awsAccessKeyId_fact }}'
+  when: awsAccessKeyId_fact is defined
+
+- name: Make sure AWS credentials are insterted in the arkcase-server.yaml file if they are defined
```
david-oc-miller: Since these two lines are already present in the server.yml template, do you really need these two `lineinfile` tasks?  If so, do you need to make sure they appear immediately after `aws.credentials:`, and do you need to make sure `aws.credentials:` itself is present?

vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -116,3 +116,9 @@ calendar:
   :\"{{ calendar_purge_option }}\",\"daysClosed\":null}}"
 acm.hazelcast:
   log.folder: "{{ root_folder }}/log/arkcase"
+
+################### AWS Credentials (aws/credentials.properties) ####################
+## AccessKeyId and SecretAccessKey must be encrypted. ##
```
david-oc-miller, line 121: Do we have to encrypt the access key and the secret access key ourselves?  If so where is the task to encrypt them?

vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -116,3 +116,9 @@ calendar:
...
+aws.credentials:
+  awsAccessKeyId: {{ awsAccessKeyId_fact }}
+  awsSecretAccessKey: {{ awsSecretAccessKey_fact }}
```
bojmil-armedia: All you need is these two lines. Nothing more, no replacements needed, as Dave said.
Also, add the awsAccessKeyId_fact and awsSecretAccessKey_fact variables in arkcase-ce-facts.yml and  arkcase_dev_facts.yml with empty string values "", to serve as a documentation of possible facts you can define.

vagrant/provisioning/roles/arkcase-app/templates/server.yml
bojmil-armedia: In would be easy to encrypt them. Just add the variable in "encrypt passwords" task and use it in the template like: "ENC({{ encrypted_aws_secret_access_key }})"

bojmil-armedia (dismissed):

dzdimov (dismissed):

bojmil-armedia (approved):

PR #261, opened 2021-02-18
Remove the decripted symmetric key at the start arkcase role

bojmil-armedia:


bojmil-armedia:
Seems like we need to encrypt passwords in multiple roles (arkcase-app, foia, oidc-config). So, I moved the deletion of the decrypted symmetric key in start-arkcase role.

PR #262, opened 2021-02-19
DSO-282 - Fix haproxy service renaming procedure

dzdimov:


dzdimov:
I have made these changes since we were getting errors that there are too many links for the file upon systemctl enable haproxy which in turn made the haproxy not being able to start upon restart

vagrant/provisioning/roles/haproxy/tasks/main.yml
```diff
@@ -447,13 +447,16 @@
     force: yes
     backup: yes
 
-- name: Change haproxy service name
+- name: Ensure that haproxy service file exists
```
fabricetriboix, line 450: I think something like "Check whether the haproxy service file exists" would better reflect what this task is doing

fabricetriboix (approved):

kikomanaskov333 (approved):

bojmil-armedia (approved):

PR #271, opened 2021-03-05
Added a default value for zylab_server_host

dzdimov:
if we don't have zylab_server_host defined the installer fails now it will pickup a default empty value

### DIFF
--- a/vagrant/provisioning/roles/arkcase-app/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-app/tasks/main.yml
@@ -451,7 +451,7 @@
   become: yes
   java_cert:
     cert_alias: "zylab_server_cert"
-    cert_url: "{{ zylab_server_host }}"
+    cert_url: "{{ zylab_server_host | default('') }}"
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -451,7 +451,7 @@
   become: yes
   java_cert:
     cert_alias: "zylab_server_cert"
-    cert_url: "{{ zylab_server_host }}"
+    cert_url: "{{ zylab_server_host | default('') }}"
```
david-oc-miller: The `java_cert` task will fail when it runs with a blank host.  Better to add a `when` clause to this task, and run it only when zylab_server_host is defined and non-blank

dzdimov: got it

david-oc-miller (approved):

david-oc-miller:
looks like it has merge conflicts now?

dzdimov:
fix was in another file (server.yml template)

PR #274, opened 2021-03-12
Fix cryptography issue by upgrading pip2 to specific version

kikomanaskov333:


vagrant/provisioning/roles/common/tasks/main.yml
```diff
@@ -29,6 +36,18 @@
       - python3
       - python3-pip
 
+- name: copy requirements.txt needed for pip2 upgrade
+  become: yes
+  copy:
+    src: requirements.txt
+    dest: /home/
```
fabricetriboix: `/home` is not a good directory for such a file. This is clearly a temporary file, so it would be much better to put it in `/tmp`.

kikomanaskov333: You're right @fabricetriboix , I hesitated between these 2 directories :)

fabricetriboix (approved):

bojmil-armedia (approved):

dzdimov (approved):

PR #278, opened 2021-03-29
Enable portal auhtentication type configuration

bojmil-armedia:


vagrant/provisioning/roles/foia/templates/arkcase-portal-server-2020.16.yaml
```diff
@@ -14,7 +14,7 @@ portal:
   arkcaseUrl: "https://{{ internal_host }}/arkcase"
   groupName: "{{ foia_portal_group }}@{{ ldap_portal_user_domain | default(ldap_user_domain) | upper}}"
   # authentication type, possible values (basic, external)
-  authenticationType: basic
+  authenticationType: {{ portal_authentication_type | default ('basic') }}
```
kikomanaskov333, line 17: Maybe the quotes are required here, "{{ portal_authentication_type | default ('basic') }}"

dzdimov: Yes, whenever we are using a variable quotes are required

bojmil-armedia: The quotes are added in the generated yaml file. And yaml requires quotes for some special characters (https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml)

By the way, It works fine like this.

PR #279, opened 2021-04-02
Added TouchNet payment variables

bojmil-armedia:


vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -133,4 +133,10 @@ zylabIntegration:
     clientSecret: "{{ zylab_clientSecret | default('') }}"
     tokenUri: "{{ zylab_tokenUri| default('') }}"
     systemUserEmail: "{{ zylab_systemUserEmail | default('') }}"
-    systemUserPassword: "{{ zylab_systemUserPassword | default('') }}"
\ No newline at end of file
+    systemUserPassword: "{{ zylab_systemUserPassword | default('') }}"
+payment:
```
david-oc-miller, line 137: do you want to always include this block?  Or include it only if some feature flag is defined, e.g. `touchnet_enabled`?

dzdimov (approved):

vagrant/provisioning/roles/arkcase-app/templates/server.yml
bojmil-armedia: I way thinking about it as well. We have these in arkcase.yaml always.
But you might be right. I'll add a flag.

bojmil-armedia: Hmm. The same goes probably for zylabIntegration part. Let's merge this one and I will fix both.

bojmil-armedia: Created https://arkcase.atlassian.net/browse/DSO-361 to fix this.

PR #281, opened 2021-04-09
Add missing snowbound fonts

cherepnalkovski:


kikomanaskov333 (approved):

bojmil-armedia (approved):

david-oc-miller:
@cherepnalkovski where did you get this new fonts file?  Is there some site or URL that we should keep checking for future updates?

cherepnalkovski:
> @cherepnalkovski where did you get this new fonts file? Is there some site or URL that we should keep checking for future updates?

https://fonts.google.com/analytics
I was looking at most popular/used fonts by google.

david-oc-miller:
Should we have a scheduled task every month or quarter, to update the font package?  If so can you write up a wiki with the steps to be taken?  Ideally we could automate this task.

PR #282, opened 2021-04-12
Update the installer pentaho-configuration role

jovestefanovski:
Replace “and” with “or” to check if the arkcase_extension_version is defined.

### DIFF
--- a/vagrant/provisioning/roles/pentaho-configuration/tasks/main.yml
+++ b/vagrant/provisioning/roles/pentaho-configuration/tasks/main.yml
@@ -349,7 +349,7 @@
 - name: format extension version
   set_fact:
     arkcase_extension_version_formatted: "{{ '-' ~ arkcase_extension_version if arkcase_extension_version != '' else '' }}"
-  when: arkcase_extension_install|bool and arkcase_extension_install_reports|bool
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/pentaho-configuration/tasks/main.yml
```diff
@@ -349,7 +349,7 @@
 - name: format extension version
   set_fact:
     arkcase_extension_version_formatted: "{{ '-' ~ arkcase_extension_version if arkcase_extension_version != '' else '' }}"
-  when: arkcase_extension_install|bool and arkcase_extension_install_reports|bool
+  when: arkcase_extension_install|bool or arkcase_extension_install_reports|bool
```
bojmil-armedia: I would just remove the whole `when` condition. Why wouldn't we format the arkcase_extension_version? Perhaps, we can check if arkcase_extension_version is defined? I don't see how this arkcase_extension_version formatting is affected by arkcase_extension_install or arkcase_extension_install_reports.

dzdimov: @jovestefanovski let's implement Bojan's suggestion and merge this

dzdimov (approved):

PR #285, opened 2021-04-14
Create a variable for arkcase-admin and change the installer to use it

jovestefanovski:
Update the foia-analytical-reports role so that it’s not using arkcase-admin as a hardcoded user but instead use {{ arkcase_admin_user }} as a variable since this user can vary from environment to environment


dzdimov (dismissed):

vagrant/provisioning/roles/httpd/templates/index.html
```diff
@@ -4,7 +4,7 @@
   </head>
   <body>
     <p>ArkCase webapp is here: <a href="/arkcase">ArkCase CE</a></p>
-    <p>ArkCase administrator account: arkcase-admin@arkcase.org, password @rKc@3e</p>
+    <p>ArkCase administrator account: {{ admin_user }}@arkcase.org, password @rKc@3e</p>
```
bojmil-armedia: The whole PR is already done. Not sure why Github shows the old version of file on develop branch.
This one here should be added. Use {{ arkcase_admin_user }}@{{ ldap_user_domain }}

vagrant/provisioning/arkcase-ce-facts.yml
```diff
@@ -37,6 +37,8 @@ samba_domain: "DC.ARKCASE-CE"
 # server_admin goes in the httpd conf file, and is the email
 # address of the arkcase-admin user
 server_admin: "david.miller@arkcase.com"
+# Not using arkcase-admin as a hardcoded user but instead use it as a variable since this user can vary from environment to environment
+admin_user: "arkcase-admin"
```
bojmil-armedia, line 41: Why don't you use the existing **arkcase_admin_user** variable? Now we have to add this new variable in all templates. Can we use arkcase_admin_user as a default?

dzdimov: I agree, we already have the arkcase_admin_user: "{{ ldap_user_prefix }}arkcase-admin"

PR #287, opened 2021-04-21
Modify the installer to change the default Alfresco password

jovestefanovski:
Currently the default Alfresco password is left when we install Alfresco and we need to modify the installer to change the password upon installation.

### DIFF
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -376,8 +376,8 @@ default_database_password: "@rKc@3S!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
-# Set Alfresco Admin password
-alfresco_admin_password: "{{ arkcase_admin_password }}" 
+# use more complicated password for Alfresco admin user before you start with the installation
+alfresco_admin_password: "admin"
 
 # Alfresco Community Edition
 alfresco_release_name: 201806-GA
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -376,6 +376,9 @@ default_database_password: "@rKc@3S!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
+# Set Alfresco Admin password
+alfresco_admin_password: "{{ arkcase_admin_password }}" 
+
 # Alfresco Community Edition
 alfresco_release_name: 201806-GA
 alfresco_build_number: "-build-00113"
--- a/vagrant/provisioning/arkcase-dev-facts.yml
+++ b/vagrant/provisioning/arkcase-dev-facts.yml
@@ -461,6 +461,9 @@ default_database_password: "@rM3d1A!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
+# Set Alfresco Admin password
+alfresco_admin_password: "{{ arkcase_admin_password }}"
+
 # Alfresco Community Edition
 alfresco_release_name: 201806-GA
 alfresco_build_number: "-build-00113"
--- a/vagrant/provisioning/arkcase-dev-facts.yml
+++ b/vagrant/provisioning/arkcase-dev-facts.yml
@@ -461,8 +461,8 @@ default_database_password: "@rM3d1A!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
-# Set Alfresco Admin password
-alfresco_admin_password: "{{ arkcase_admin_password }}"
+# use more complicated password for Alfresco admin user before you start with the installation
+alfresco_admin_password: "admin"
 
 # Alfresco Community Edition
 alfresco_release_name: 201806-GA
--- a/vagrant/provisioning/roles/alfresco/files/passencrypt.py
+++ b/vagrant/provisioning/roles/alfresco/files/passencrypt.py
@@ -1,6 +1,0 @@
-#!/usr/bin/env python
-import sys
-import hashlib
-passwd = sys.argv[1]
-passwd.encode('utf-16le')
-print hashlib.new('md4', passwd.encode('utf-16le')).hexdigest()
--- a/vagrant/provisioning/roles/alfresco/files/passencrypt.py
+++ b/vagrant/provisioning/roles/alfresco/files/passencrypt.py
@@ -1,0 +1,6 @@
+#!/usr/bin/env python
+import sys
+import hashlib
+passwd = sys.argv[1]
+passwd.encode('utf-16le')
+print hashlib.new('md4', passwd.encode('utf-16le')).hexdigest()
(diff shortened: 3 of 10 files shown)


vagrant/provisioning/roles/alfresco/files/passencrypt.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
+import sys
+import hashlib
+passwd = sys.argv[1]
+passwd.encode('utf-16le')
```
fabricetriboix: This line does nothing

vagrant/provisioning/arkcase-ce-facts.yml
```diff
@@ -376,6 +376,9 @@ default_database_password: "@rKc@3S!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
+# Set Alfresco Admin password
+alfresco_admin_password: "{{ arkcase_admin_password }}" 
+
```
bojmil-armedia, line 381: Add this to arkcase-dev-facts.yml

vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -472,8 +472,25 @@
...
+
+- name: Run Python script to generate the hash
+  script: "passencrypt.py {{ alfresco_admin_password }}"
+  register: pass_hash
+
+- name: Execute required update querry to change default alfresco password
+  command: >
+    mysql -u root -D alfresco -e "UPDATE alf_node_properties SET string_value='{{ pass_hash.stdout | trim }}' WHERE node_id=(SELECT anp1.node_id FROM alf_node_properties anp1 INNER JOIN alf_qname aq1 ON aq1.id = anp1.qname_id INNER JOIN alf_node_properties anp2 ON anp2.node_id = anp1.node_id INNER JOIN alf_qname aq2 ON aq2.id = anp2.qname_id WHERE aq1.local_name = 'password' AND aq2.local_name = 'username' AND anp2.string_value = 'admin') and qname_id=(SELECT anp1.qname_id FROM alf_node_properties anp1 INNER JOIN alf_qname aq1 ON aq1.id = anp1.qname_id INNER JOIN alf_node_properties anp2 ON anp2.node_id = anp1.node_id INNER JOIN alf_qname aq2 ON aq2.id = anp2.qname_id WHERE aq1.local_name = 'password' AND aq2.local_name = 'username' AND anp2.string_value = 'admin');"
+
+- name: Check Alfresco if active
```
fabricetriboix: Typo: "if" -> "is"

fabricetriboix (changes requested):

bojmil-armedia (commented):
We should also update the password in `camel-config-alfresco-cmis.properties` in `arkcase-app/tasks/main.yml`

vagrant/provisioning/roles/alfresco/files/passencrypt.py
marjan4jc: Why this script is called passencrypt.py when only encode the password and then create a hash from it, and why as a hash algorithm md4 is used, it is considered insecure since 1990, md5 is considered insecure for 20 years now. A better choice as a hash alg would be one of the SHA-2 variants.

bojmil-armedia: I agree with the file naming.
Anyway, the MD4 algorithm is used by Alfresco, we cannot change that.

marjan4jc: @bojmil-armedia , this can be changed. 

With Bcrypt, the hashing algorithm (also called an encoder) can be configured by setting the system.preferred.password.encoding property in the alfresco-global.properties file. The supported values for this property are:

md4
sha256
bcrypt10
If you provide a different value, the repository won’t start.

To maintain backwards compatibility with previous versions, the default setting for this property is:

system.preferred.password.encoding=md4


Here is the link:
https://docs.alfresco.com/content-services/6.1/admin/security/#bcryptoverview

SHA-256 is not best choise but much much better then MD4.  I do not know about bcrypt10, someone can investigate if that is the best option.

bojmil-armedia: Wow. I guess, they keep MD4 as default for backward compatibility or something...

david-oc-miller: Let's try with bcrypt10, it should work according to Alfresco docs, and it should be best for new installations.  For current installations, real user passwords are in LDAP, not in Alfresco, so we only have to upgrade admin to the new hashing.

bojmil-armedia: The admin password is only in the Alfresco database. We don't need to change anything. Except, if we want to change the password. But that can be done manually.

david-oc-miller: If we change the hash algorithm for existing installations, I suppose we need to update the Alfresco admin password on those sites, but we can experiment locally to see how that works.

vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -517,6 +510,25 @@
     name: alfresco
     state: started
 
+- name: install bcrypt for hashing
+  become: yes
+  pip:
```
david-oc-miller, line 515: Do you have to specify pip3 somehow here?  On some sites if pip2 is installed, I've seen an error when running plain pip, without specifying the pip version.

vagrant/provisioning/arkcase-ce-facts.yml
kikomanaskov333: @jovestefanovski , change this variable to "admin" since the default pass is admin on dev environments

vagrant/provisioning/arkcase-dev-facts.yml
```diff
@@ -461,6 +461,9 @@ default_database_password: "@rM3d1A!"
 tomcat_major_version: 9
 tomcat_version: 9.0.22
 
+# Set Alfresco Admin password
+alfresco_admin_password: "{{ arkcase_admin_password }}"
```
kikomanaskov333: @jovestefanovski , the same here

vagrant/provisioning/roles/alfresco/tasks/main.yml
```diff
@@ -500,6 +510,26 @@
...
+    executable: pip3
+    name: bcrypt
+
+- name: run Python script to generate the hash
+  script: "passencode.py {{ alfresco_admin_password }}"
```
kikomanaskov333, line 520: Set 'admin' as default if this variable alfresco_admin_password is not defined

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -23,8 +23,8 @@
         value: "{{ default_database_password }}"
       - name: default_user
         value: "{{ default_user_password }}"
-      - name: default_alfresco
-        value: admin
+      - name: alfresco_admin_password
+        value: "{{ alfresco_admin_password }}"
```
kikomanaskov333, line 27: @jovestefanovski, same here :)

PR #291, opened 2021-05-19
Create cronjob in the installer to remove all files older than one week

jovestefanovski:
The fileserver folder (arkcase/data/arkcase-home/fileserver) is used to transfer files to and from ActiveMQ.  Files are added there, but never removed.

### DIFF
--- a/vagrant/provisioning/roles/common/tasks/main.yml
+++ b/vagrant/provisioning/roles/common/tasks/main.yml
@@ -111,3 +111,14 @@
 - name: Set default locale to en_US.utf8
   become: yes
   command: localectl set-locale LANG=en_US.utf8
+
+- name: Run CRON job every day at 1.05 to remove files older than 7 days
+  become: yes
+  become_method: sudo
+  cron:
+    name: "remove_files"
+    user: "root"
+    hour: "1"
+    minute: "5"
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/common/tasks/main.yml
```diff
@@ -111,3 +111,14 @@
...
+
+- name: Run CRON job every day at 1.05 to remove files older than 7 days
+  become: yes
+  become_method: sudo
+  cron:
+    name: "remove_files"
+    user: "root"
+    hour: "1"
+    minute: "5"
+    job: 'find {{ root_folder }}/data/arkcase-home/fileserver -mtime +6 -type f -exec rm {} \;'
```
dzdimov, line 123: @jovestefanovski is +6 actually for files older than 6 or 7 days?

dzdimov (approved):

vagrant/provisioning/roles/common/tasks/main.yml
david-oc-miller: The job itself looks great, thanks.

It should be moved to an ArkCase role (should not be in common) since in multi-node installations, only the ArkCase node would need this job (the Snowbound, ActiveMQ, Solr, nodes wouldn't have this folder and the job would fail).

PR #292, opened 2021-05-20
New ansible roles that support backup 

kikomanaskov333:


vagrant/provisioning/roles/backup_data/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
...
+    name: snowbound
+    state: stopped    
+    
+- name: backup data folders 
+  become: yes
+  file:
+    path: "{{ root_folder }}/backups/data"
+    state: present
+
+- name: Solr data backuping
```
fabricetriboix, line 25: I don't think "backuping" is a word in English... Maybe "Backing up Solr data" would be more appropriate?
Ditto for the other ones

vagrant/provisioning/roles/backup_data/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
...
+  become: yes
+  copy:
+    src: "{{ root_folder }}/data/solr"
+    dest: "{{ root_folder }}/backups/data/solr-{{ ansible_date_time.iso8601 }}"
+    directory_mode: yes
+    owner: solr
+    group: solr
+    backup: yes
+
+- name: Alfresco data backuping
```
fabricetriboix, line 35: We need to be clear here. This "backup" is not enough to restore Alfresco. I am not foo familiar with Alfresco, but I suppose a backup of the database would be required as well, and may of configuration files. A "backup" usually implies that you can restore a state from it.

I would suggest to rename this task "Backing up Alfresco files", to make it more clear what is being done here.

vagrant/provisioning/roles/backup_jar_war/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    state: stopped
+  when: foia_portal_context is defined
+
+- name: backup data folders 
+  become: yes
+  file:
+    path: "{{ root_folder }}/backups/webapps"
+    state: present
+
+- name: Config-Server jar backuping
```
fabricetriboix, line 26: Why should jar/war files be backed up? They can be restored using the same installation process that deployed them initially.

bojmil-armedia: The versions might have changed. We don't have any place where we keep the old version.

bojmil-armedia:
We should backup the pentaho database. It keeps privileges, users, and deployed reports (hibernate, jackrabbit, quartz).

PR #304, opened 2021-06-22
Add facts which are required during the vagrant provisioning process

jovestefanovski:
Include arkcase_version facts in arkcase-dev-facts.yml


david-oc-miller:
I just hit this issue locally, thanks for checking in the fix.

david-oc-miller (approved):

PR #308, opened 2021-07-02
add mitogen support to tower

jovestefanovski:


david-oc-miller:
What is mitogen?  Why do we need to fork Mitogen right into the installer code?

jovestefanovski:
We just wanted to improve the speed of role execution when doing deployments through awx tower and we were testing, but mitogen is not supported with the ansible version we are using right now. The commit is reverted.

PR #335, opened 2021-09-27
DSO-449  Change default compliance for Alfresco RM site to Standard site

viktorijakostovska:
Added new parameter for rm site type and compliance.

### DIFF
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -381,6 +381,8 @@ alfresco_release_name: 201806-GA
 alfresco_build_number: "-build-00113"
 alfresco_content_services_version: 6.0.7-ga
 alfresco_records_management_version: 3.0.a
+# Alfresco RM site type for dod5015 is '{http://www.alfresco.org/model/dod5015/1.0}site' and for standard type is '{http://www.alfresco.org/model/recordsmanagement/1.0}rmsite'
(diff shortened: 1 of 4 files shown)


vagrant/provisioning/arkcase-ce-facts.yml
```diff
@@ -381,6 +381,8 @@ alfresco_release_name: 201806-GA
 alfresco_build_number: "-build-00113"
 alfresco_content_services_version: 6.0.7-ga
 alfresco_records_management_version: 3.0.a
+# Alfresco RM site type for dod5015 is '{http://www.alfresco.org/model/dod5015/1.0}site' and for standard type is '{http://www.alfresco.org/model/recordsmanagement/1.0}rmsite'
+alfresco_records_management_site_type: "{http://www.alfresco.org/model/dod5015/1.0}site"
```
bojmil-armedia, line 385: @dzdimov @jovestefanovski when installing a new environment from now on, let's use the standard type  
```
alfresco_records_management_site_type: "{http://www.alfresco.org/model/recordsmanagement/1.0}rmsite"  
```

The DOD site has some complications for users that do not really require DOD type.

bojmil-armedia (approved):

vagrant/provisioning/arkcase-ce-facts.yml
david-oc-miller: I agree, this change looks good, if you can reverse the default values (default to be {http://www.alfresco.org/model/recordsmanagement/1.0}rmsite)

PR #336, opened 2021-09-30
Support Vault configuration within the installer

jovestefanovski:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -1035,3 +1035,26 @@
...
+- name: replace values in "conf.yml"
+  become: yes
+  become_user: arkcase
+  replace:
+    backup: yes
+    path: /home/arkcase/.arkcase/acm/conf.yml
+    regexp: "{{ item.regexp }}"
+    replace: "{{ item.replace }}"
+  loop:
+    - regexp: "(  host: )(.*)"
```
bojmil-armedia, line 1047: Let's use yaml cli tool for these replacements.

bojmil-armedia:
Used jq with better yaml support.

PR #338, opened 2021-10-01
Ignore media metadata properties detected by tika

cherepnalkovski:
Create new dynamic field with prefix "ignored_" and set type=ignored
Map media_white_point and media_black_point properties detected by tika with the new ignored prefix.

This is known Solr issue, there is a [ticket ](https://issues.apache.org/jira/browse/SOLR-8017)on their Jira and the only way to solve it is to ignore the property detected by Solr-Tika service.

On environments with Solr in standalone mode we should only update those files.
On environments with Solr in cloud mode, we should update those files and register them on zookeeper.

./zkcli.sh -cmd upconfig -zkhost 127.0.0.1:2181 -confname arkcaseConfig -confdir /path to the folder with the config files.

### DIFF
--- a/vagrant/provisioning/roles/solr/files/solrconfig.xml
+++ b/vagrant/provisioning/roles/solr/files/solrconfig.xml
@@ -801,6 +801,8 @@
       <str name="uprefix">ignored_</str>
       <str name="captureAttr">true</str>
 
+      <str name="fmap.media_white_point">ignored_</str>
+      <str name="fmap.media_black_point">ignored_</str>
       <str name="fmap.meta">ignored_</str>
       <str name="fmap.content">content</str>
       <str name="fmap.a">links</str>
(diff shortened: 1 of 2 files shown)


bojmil-armedia (approved):

vagrant/provisioning/roles/solr/files/solrconfig.xml
```diff
@@ -801,6 +801,8 @@
       <str name="uprefix">ignored_</str>
       <str name="captureAttr">true</str>
 
+      <str name="fmap.media_white_point">ignored_</str>
```
david-oc-miller, line 804: Maybe include a link to the Solr issue here, so sometime in future we can see if the issue is fixed and possibly remove these mappings ... although we don't really care about these fields anyway.  But a ticket link would explain why we ignore these fields.

PR #340, opened 2021-10-18
Installer should set -Dspring.profiles.active=ldap on Core environments

jovestefanovski:
We set active profiles for foia and privacy, but not for core variant.

### DIFF
--- a/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
@@ -143,3 +143,23 @@
     url: "https://github.com/mikefarah/yq/releases/download/v4.13.2/yq_linux_amd64"
     dest: /usr/bin/yq
     mode: '0755'
+
+- name: read current setenv.sh
+  become: yes
+  become_user: arkcase
+  command: cat {{ root_folder }}/app/arkcase/bin/setenv.sh
+  register: setenv_contents
+  changed_when: false
+
+- name: ensure ldap is in the profile (no other profiles already exist)
+  become: yes
+  become_user: arkcase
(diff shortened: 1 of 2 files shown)


dzdimov (dismissed):

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -143,3 +143,23 @@
...
+  become_user: arkcase
+  blockinfile:
+    path: "{{ root_folder }}/app/arkcase/bin/setenv.sh"
+    insertbefore: EOF
+    marker: "# {mark} ANSIBLE MANAGED BLOCK - Core"
+    block: |
+      export JAVA_OPTS="$JAVA_OPTS -Dspring.profiles.active=ldap"
+  when: 
+    - '"JAVA_OPTS -Dspring.profiles.active=ldap" not in setenv_contents.stdout'
+    - arkcase_configuration_classifier == "core"
```
bojmil-armedia: We don't need this one. For an extension the arkcase_configuration_classifier would be another one. We can add this one if there is no "-Dspring.profiles.active" string in the file. 
If we have a foia installation that one will add a second line export JAVA_OPTS="$JAVA_OPTS -Dspring.profiles.active=foia-extension,ldap" and we need to check what will happen.

bojmil-armedia: Maybe add this at the end somewhere. If no role set this one up, add it.

bojmil-armedia: arkcase-post-upgrade ?

bojmil-armedia (approved):

PR #347, opened 2021-11-08
Introduce Kafka to Arkcase

bojmil-armedia:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -258,13 +258,23 @@
...
 
-- name: copy config-server configuration file
+- name: copy config-server configuration file before version 2021.04
   become: yes
   become_user: arkcase
   template:
     src: config-server.conf
     dest: "{{ root_folder }}/app/config-server/config-server.conf"
     force: yes
+  when: arkcase_version != "" and arkcase_version is version('2021.04', '<')
```
bojmil-armedia, line 283: This might need to be changed to 2022.01. We'll see...

bojmil-armedia: It will work anyway

PR #355, opened 2021-12-15
Fix for CVE-2021-44228

dzdimov:
We have added the fix for the installer for previous versions of ArkCase where we don't have the Apache Log4j updated version that is not vulnerable to this vulnerability


jovestefanovski (approved):
this looks good

PR #357, opened 2021-12-22
Draft: Remove config-server dependencies for older Arkcase versions

jovestefanovski:
This is initially created for MDSAP and it can be used in future where we have to deploy older Arkcase versions through the installer.


vagrant/provisioning/roles/start-arkcase/tasks/main.yml
```diff
@@ -4,19 +4,6 @@
     path: "{{ root_folder }}/common/symmetricKey.decrypted"
     state: absent
 
-- name: start config server
-  become: yes
-  systemd:
-    daemon_reload: true
-    name: "config-server"
-    state: started
```
bojmil-armedia, line 12: Are we starting it somewhere else? Why don't you just put a when condition on arkcase_version?

bojmil-armedia: It should not run prior arkcase version 3.3.1

david-oc-miller:
@jovestefanovski , @dzdimov , do we still need remove config-server dependencies for older ArkCase versions?  It's hard to see the actual change now since some of the Alfresco 7 / PostgreSQL changes are layered on top.

jovestefanovski:
This PR was specifically for MDSAP with PR source DSO-480. And this shouldn't be merged into develop.
With just using this code with single run we were able to deploy older ArkCase version with PostgreSQL to suit MDSAP.

david-oc-miller:
Thanks @jovestefanovski ... so we can close this MR without merging?

PR #365, opened 2022-01-07
RHEL 8 support 

david-oc-miller:
Common, samba, common


david-oc-miller:
merged this one since I never got any comments, and also I don't see any breaking changes.

PR #368, opened 2022-01-11
Fix: AWS AMI Startup script cannot contact ldap server

jovestefanovski:
During the execution of the startup script runs into error: Cannot contact ldap server.
Missing name reference to ip in the hosts file.

### DIFF
--- a/vagrant/provisioning/roles/common/tasks/main.yml
+++ b/vagrant/provisioning/roles/common/tasks/main.yml
@@ -97,7 +97,7 @@
     line: "{{ item }}"
   loop:
     - "{{ arkcase_host_address | default('127.0.0.1') }} arkcase-host"
-    - "{{ ansible_default_ipv4.address }} {{ internal_host }}"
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/common/tasks/main.yml
```diff
@@ -97,7 +97,7 @@
     line: "{{ item }}"
   loop:
     - "{{ arkcase_host_address | default('127.0.0.1') }} arkcase-host"
-    - "{{ ansible_default_ipv4.address }} {{ internal_host }}"
+    - "{{ ansible_default_ipv4.address | default(ansible_all_ipv4_addresses[0]) }} {{ internal_host }}"
```
david-oc-miller, line 100: Interesting that `ansible_default_ipv4.address` is not identical to `ansible_all_ipv4_addresses[0]` ... good catch here.

david-oc-miller (approved):

PR #369, opened 2022-01-12
Fix: Restored deletion of localhost to point to internal host, since …

jovestefanovski:
…AWS AMI gets wrong IP during the building process


david-oc-miller:
Using 127.0.0.1 for internal_host broke Solr, that is why I changed it to a real IP address. With e.g. "127.0.0.1 arkcase-ce.local", ArkCase could not connect to Solr due to the haproxy configuration.  But it seems fine with the real IP address.

david-oc-miller:
I'll see if Solr works OK with the GitHub release, if we have both the 127.0.0.1 and the real IP address pointing to arkcase-ce.local.

jovestefanovski:
It should work like that, if not we can change slightly the startup script to insert what we want in hosts file at first.

dzdimov (approved):

dzdimov:
Merging this to test if solr will work with both entries in the hosts file

PR #375, opened 2022-01-20
DSO-529 Fix nodejs installation

dzdimov:
Currently epel-release does not contain nodejs (nodejs package not found) so we are resorting to a different approach to installing nodejs-by using nvm

### DIFF
--- a/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
@@ -63,6 +63,18 @@
       owner: root
       url: https://dl.yarnpkg.com/rpm/yarn.repo
 
+- name: ensure the nvm install script is downloaded
+  command: sudo curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+
+- name: Export Vars to run npm without the need to logout
+  command: export NVM_DIR="$HOME/.nvm" [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
+
+- name: Install nodejs using nvm
+  command: sudo nvm install node
+
+- name: Make sure correct (6.17.1) version of npm is installed
+  command: sudo nvm install 6.17.1
+
 - name: required packages to run ArkCase webapp
   become: yes
   yum:
@@ -70,7 +82,6 @@
     name:
       - yarn
       - npm
-      - nodejs
 
 - name: install python2 (needed on RHEL8)
   become: yes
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -55,27 +55,37 @@
       vars:
         item: arkcase
 
+- name: "Check if node is installed"
+  become: yes
+  shell: "node --version"
+  register: is_present
```
david-oc-miller, line 61: `is_present` isn't clear ... what is present?  Maybe `node_is_present` or `node_version`

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -55,27 +55,37 @@
...
   shell: "curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash"
+  when: is_present.stdout == ""
 
 - name: install node, npm and set version
   become: yes
   become_user: root
   shell: "source /root/.nvm/nvm.sh && nvm install 6.17.1 && nvm alias default v6.17.1"
   args:
     executable: /bin/bash
+  when: is_present.stdout == ""
```
david-oc-miller, line 75: sooner or later we'll have to extract the nvm and node versions into facts where they can easily be changed.

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -55,27 +55,37 @@
...
 
 - name: install the module 'yarn' with npm required by ArkCase
   become: yes
   npm:
     name: yarn
     global: yes
     state: present
+  when: is_present.stdout == ""
```
david-oc-miller, line 83: Just because node was not installed previously, doesn't mean yarn was also not installed, but maybe it doesn't hurt to install yarn again even if it is already there.

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -55,27 +55,37 @@
...
 
 - name: ensure node, npm and yarn are available for all users
   become: yes
   shell: "n=$(which node);n=${n%/bin/node}; chmod -R 755 $n/bin/*; sudo cp -r $n/{bin,lib,share} /usr/local"
+  when: is_present.stdout == ""
```
david-oc-miller, line 88: Does the wildcard affect only `node` or could it possibly affect other files that are not part of the ArkCase installation?

david-oc-miller:
If we still need this change, can you address the code review comments?

jovestefanovski:
We don't need this change, since epel-release NodeJS repos were not reachable for couple of days, and now are back.
Our intent was to not depend by the epel-release node package and use nvm, which was looking very promising, but not when you try to deploy it with ansible.
We tried to get it working flawlessly but without luck. 
It's good to have something as a backup solution, but this is not the best one.
This can be closed.

PR #379, opened 2022-01-27
DSO-538 httpd cannot start after upgrade

jovestefanovski:


vagrant/provisioning/roles/microservices/tasks/main.yaml
```diff
@@ -47,6 +47,15 @@
   when: enable_kafka is defined and enable_kafka == "yes"
   register: control_center_httpd 
 
+- name: replace original '00-base.conf' file with upgraded one to prevent conflict during start
+  copy:
+    remote_src: yes
+    force: yes
+    backup: yes
+    src: /etc/httpd/conf.modules.d/00-base.conf.rpmnew
+    dest: /etc/httpd/conf.modules.d/00-base.conf
```
armedia-customer-service, line 61: Is the file `00-base.conf.rpmnew` guaranteed to exist, if `httpd_upgrade is changed or control_center_httpd is changed`?  What happens on a new install, where the initial httpd version was the target version, and hasn't been upgraded since it is already the newest?

jovestefanovski: I did change the approach for this scenario, now if the file doesn't exists, the task for copying will be skipped.

dzdimov (approved):

PR #386, opened 2022-02-01
Fix Kafka for portal

bojmil-armedia:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -848,6 +848,15 @@
     src: server.yml
     dest: /home/arkcase/.arkcase/acm/acm-config-server-repo/arkcase-server.yaml
 
+- name: server-specific configuration for arkcase-commons.yml for version 2021.04
```
armedia-customer-service, line 851: `2021.04` ???  should it be '2022.01` ?

bojmil-armedia: It works the same. We started with 2021.04 and didn't want to change everywhere.

PR #388, opened 2022-02-10
Deploy ArkCase microservices

bojmil-armedia:


vagrant/provisioning/arkcase-all.yml
```diff
@@ -70,6 +70,39 @@
       tags: [arkcase-prerequisites, core, arkcase, upgrade, marketplace]
     - role: arkcase-app
       tags: [arkcase-app, core, arkcase, upgrade, marketplace]
+    - role: arkcase-api-gateway
+      tags: [core, upgrade, arkcase-api-gateway, arkcase-microservices, marketplace]
```
david-oc-miller, line 98: Is "yes" preferred over a boolean type?  e.g. `when: enable_kafka is defined and enable_kafka|bool`?

david-oc-miller: Also, maybe `enable_microservices` versus `enable_kafka`?

vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
```diff
@@ -1,3 +1,8 @@
 arkcase:
   kafka:
     bootstrapAddress: {{ kafka_host }}:{{ kafka_port | default(9092) }}
+    schemaRegistryAddress: https://{{ internal_host }}/schema-registry
```
david-oc-miller: Will the schema registry always be on the same host as this commons-server?  Or would it be on the `kafka_host` or some other host?

vagrant/provisioning/roles/arkcase-file-management/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+- name: deploy arkcase-file-management
+  include_tasks: "{{ role_path }}/../arkcase-microservice/tasks/deploy_microservice.yml"
+  vars:
+    microservice:
+      name: arkcase-file-management
+      version: "{{ arkcase_file_management_version | default(arkcase_version) }}"
+      java_opts: "-Dserver.port=9068 -Dspring.cloud.config.uri=https://{{ config_server_host | default(internal_host) }}/config -Dspring.profiles.include=server,runtime -Djava.net.preferIPv4Stack=true -Djavax.net.ssl.trustStorePassword={{ java_trust_store_pass }} -Djavax.net.ssl.keyStore={{ java_key_store }} -Djavax.net.ssl.trustStore={{ java_trust_store}} -Djavax.net.ssl.keyStorePassword={{ java_key_store_pass }} -Dlogging.file.name={{ root_folder }}/log/arkcase-file-management/arkcase-file-management.log -Djava.io.tmpdir={{ root_folder }}/data/arkcase-file-management/tmp"
```
david-oc-miller: Can we construct the `logging.file.name` and `java.io.tmpdir` parameters in the `deploy_microservice` task, to avoid a situation where somebody changes the `name` without changing these paths?

vagrant/provisioning/arkcase-all.yml
bojmil-armedia: We have enable_kafka variable, so I reused it here. My thinking was that without Kafka you cannot have microservices.
The existing variable was "yes" and didn't wan't to change it. I prefer boolean true/false. Having said this, I can change this, because it won't really change much at this point. So, I'll introduce enable_microservices boolean variable and get rid of enable_kafka variable.

vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
bojmil-armedia: I will introduce variables for each microservice like arkcase_schema_host, arkcase_api_gateway etc., and use internal_host as default

vagrant/provisioning/roles/arkcase-file-management/tasks/main.yml
bojmil-armedia: I can append them automatically. I was thinking about more freedom of what you can put here, but I don't see a downside to setting these two by default.

david-oc-miller: Well, if you create the folders in this task, then all is well too, but I see the deploy task creates the folders based on the `name`.

vagrant/provisioning/roles/arkcase-microservice/tasks/deploy_microservice.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: copy {{ microservice.name }} configuration file
+  become: yes
+  become_user: arkcase
+  template:
+    src: ../templates/microservice.conf
+    dest: "{{ root_folder }}/app/{{ microservice.name }}/{{ microservice.name }}.conf"
+    force: yes
+
+- name: "enable {{ microservice.name }}"
```
david-oc-miller, line 80: Is TLS enabled on the microservices?  I think we still don't have TLS on config-server.  TLS on all endpoints is a constraint / requirement for FedRAMP.

bojmil-armedia: No TLS unfortunately on any of the microservices (including config server). We will need to work on that in another ticket,

vagrant/provisioning/roles/arkcase-file-management/tasks/main.yml
bojmil-armedia: I meant, I agree.

vagrant/provisioning/roles/showcase/templates/startup.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
```
david-oc-miller, line 2: When we remove the Angular resource copier from ArkCase, then the ArkCase deployment won't require yarn, node, or npm, and we would remove those tasks from the ArkCase installer.  When that happens, we can add them to this showcase installer... but should be good for now, if we assume showcase is always installed with ArkCase.

vagrant/provisioning/roles/arkcase-microservice/tasks/deploy_microservice.yml
david-oc-miller: Yep, we need a different ticket for TLS endpoints.  If the services use messaging exclusively, they wouldn't actually need to listen on any port, so no TLS needed; but I guess at least the queries may always have an HTTP endpoint.

david-oc-miller:
It looks good to me now, but GitHub says the two facts files have conflicts.  Can accept after conflict resolution.

david-oc-miller (dismissed):

PR #392, opened 2022-02-24
ACS-508 Fix for solr log4j vulnerabilities

jovestefanovski:
solr prometheus-exporter and solr server are using vulnerable log4j libraries, which needs to be changed with the latest patched.

### DIFF
--- a/vagrant/provisioning/roles/solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/solr/tasks/main.yml
@@ -98,3 +98,83 @@
 - name: Expose required ports for solr
   include_tasks: firewall.yml
   when: enable_firewall is undefined or enable_firewall
+
+- name: find prometheus-exporter log4j jars that are vulnerable
+  find:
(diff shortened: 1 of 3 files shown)


dzdimov (approved):

vagrant/provisioning/roles/solr/tasks/main.yml
```diff
@@ -97,4 +97,84 @@
...
 - name: Expose required ports for solr
   include_tasks: firewall.yml
-  when: enable_firewall is undefined or enable_firewall
\ No newline at end of file
+  when: enable_firewall is undefined or enable_firewall
+
+- name: find prometheus-exporter log4j jars that are vulnerable
+  find:
+    paths: "{{ root_folder }}/app/solr/contrib/prometheus-exporter/lib"
+    patterns: "log4j*"
```
david-oc-miller, line 105: This change is not idempotent, since you are removing files `log4j*` and replacing them with files that also match the pattern `log4j*`.  As far as I can tell every time you run this play, the log4j libraries will be removed and replaced, even with the same version every time.

PR #396, opened 2022-03-08
Get AWX to install needed modules

david-oc-miller:


david-oc-miller:
Unfortunately I have no way to test this, except to run it from AWX, so merging now.

PR #406, opened 2022-03-15
ACS-457 Add activemq username and password in alfresco-global.properties

martinbozinovski:
Add activemq credentials in alfresco-global.properties. 
*Note: how this change will affect something? I honestly don't know if this changes anything. Without it, it is working fine. With this changes, it is working same. What is the outcome that we want here? Please elaborate someone if it is familiar with this issue..


bojmil-armedia (approved):

david-oc-miller (approved):

david-oc-miller:
@martinbozinovski , when we want Alfresco to use a non-default account for accessing ActiveMQ, this change will allow it.  For default installations, there is no change.

martinbozinovski:
Thanks for the info @david-oc-miller

PR #407, opened 2022-03-22
ACS-544 Remove alfresco-search-services role from dev playbook

martinbozinovski:
Remove alfresco-search-services from dev playbook.


dzdimov (approved):

dzdimov:
Let's investigate what's causing Alfresco-search-services to fail during startup as we will probably need it later.

PR #408, opened 2022-03-22
Deploy ArkCase Sidecar and Install Open Office

bojmil-armedia:


martinbozinovski (approved):

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -5,3 +5,59 @@
...
+- include_tasks: "{{ role_path }}/../common/tasks/download.yml"
+  loop:
+    - name: Open Office
+      dest: "/tmp/Apache_OpenOffice_{{ open_office_version | default('4.1.11') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
+      owner: root
+      url: "https://sourceforge.net/projects/openofficeorg.mirror/files/{{ open_office_version | default('4.1.11') }}/binaries/en-US/Apache_OpenOffice_{{ open_office_version | default('4.1.11') }}_Linux_x86-64_install-rpm_en-US.tar.gz --no-check-certificate"
+  when: open_office_latest_version is not defined or not open_office_latest_version
+
+- name: unpack Open Office RPM
+  become: yes
```
david-oc-miller, line 43: make sure to use `creates` property hear, so that you only unarchive OpenOffice if it hasn't already been unarchived

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -5,3 +5,59 @@
...
+
+- name: install Open Office
+  become: yes
+  command: 
+    cmd: rpm -Uvih /tmp/en-US/RPMS/*.rpm
+    warn: false
+  when: open_office_latest_version is not defined or not open_office_latest_version
+
+- name: remove tmp files
+  become: yes
```
david-oc-miller, line 58: Hmm... by removing the install files, now you can't tell whether OpenOffice was already unarchived.  Still, somehow we want to avoid the download, unarchive, and install, if the target version has already been installed.

bojmil-armedia: The first 3 steps check the installed version:
- Check Open Office installed
- Check Open Office version
- Set open_office_latest_version fact

 And all following steps use when condition (openoffice_bin.stat.exists and not open_office_latest_version). This way if you have the same version nothing is executed. If you have an older version it will be updated.

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -5,3 +5,59 @@
...
+- name: Check Open Office installed
+  become: yes
+  stat:
+    path: "/opt/openoffice4/program/soffice"
+  register: openoffice_bin 
+
+- name: Check Open Office version
+  become: yes
+  command: "/opt/openoffice4/program/soffice -h  2>&1 | head -n1"
+  register: openoffice_old_version
```
david-oc-miller, line 18: You might want the `when: openoffice_bin.stat.exists` guard here too, for clarity, even though on Ubuntu the exit code is 0 even when the executable doesn't exist (so the play won't fail).

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
david-oc-miller: I see now, very nice.

PR #409, opened 2022-03-24
ACS-546 Change the install logic for node,npm,yarn

martinbozinovski:
We had few difficulties with this packages and their versions on several environment. Hopping that this will solve some issues.

### DIFF
--- a/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
+++ b/vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
@@ -64,14 +64,47 @@
       owner: root
       url: https://dl.yarnpkg.com/rpm/yarn.repo
 
-- name: required packages to run ArkCase webapp
+- name: Get yarn version
+  become: yes
+  shell: "yarn -v"
+  register: yarn_version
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -64,14 +64,47 @@
       owner: root
       url: https://dl.yarnpkg.com/rpm/yarn.repo
 
-- name: required packages to run ArkCase webapp
+- name: Get yarn version
+  become: yes
+  shell: "yarn -v"
+  register: yarn_version
```
david-oc-miller, line 70: I guess the purpose of this merge request is to avoid installing the operating system RPM package, if yarn or node or npm was installed in some other way.

david-oc-miller (approved):

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
martinbozinovski: Exactly.

dzdimov (approved):

PR #412, opened 2022-03-28
ACS-548 Add heap size options to java_opts for victoriapd

martinbozinovski:
Add initial and maximum heap size for jvm.

### DIFF
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -614,7 +614,3 @@ arkcase_portal_extension_formatted: "{{ '-' ~ arkcase_portal_extension if arkcas
 deploy_arkcase_portal: false
 
 open_office_version: "4.1.11"
-
-# ArkCase memory size
-arkcase_mem_max: ""
-arkcase_mem_min: ""
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -614,3 +614,7 @@ arkcase_portal_extension_formatted: "{{ '-' ~ arkcase_portal_extension if arkcas
 deploy_arkcase_portal: false
 
 open_office_version: "4.1.11"
+
+# ArkCase memory size
+arkcase_mem_max: ""
+arkcase_mem_min: ""
--- a/vagrant/provisioning/roles/tomcat/tasks/main.yml
+++ b/vagrant/provisioning/roles/tomcat/tasks/main.yml
@@ -168,6 +168,19 @@
         marker: "# {mark} ANSIBLE MANAGED BLOCK - JAVA_OPTS"
         block: |
           export JAVA_OPTS="-Djava.net.preferIPv4Stack=true -Djavax.net.ssl.trustStorePassword={{ java_trust_store_pass }} -Djavax.net.ssl.trustStore={{ java_trust_store }} -Dlog4j.formatMsgNoLookups=true"
+      when: arkcase_extension_id != "victoria-pd" 
+    - name: set JAVA_OPTS
+      become: yes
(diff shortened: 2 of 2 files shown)


vagrant/provisioning/roles/tomcat/tasks/main.yml
```diff
@@ -168,6 +168,19 @@
         marker: "# {mark} ANSIBLE MANAGED BLOCK - JAVA_OPTS"
         block: |
           export JAVA_OPTS="-Djava.net.preferIPv4Stack=true -Djavax.net.ssl.trustStorePassword={{ java_trust_store_pass }} -Djavax.net.ssl.trustStore={{ java_trust_store }} -Dlog4j.formatMsgNoLookups=true"
+      when: arkcase_extension_id != "victoria-pd" 
+    - name: set JAVA_OPTS
+      become: yes
+      become_user: "{{ tc.user }}"
```
david-oc-miller: Do we really think that victoria-pd is the only customer who will ever want custom Java memory options for ArkCase?  Or do you suppose we'll keep some lookup of customer-to-memory options?

In all other such situations we just define a config item directly: `arkcase_mem_max=8g`, `arkcase_mem_min=1g`.  Then in the task we use the default filter, in case these new facts are not defined:

```
export JAVA_OPTS="-Xms{{ arkcase_mem_min | default('1g') }} -Xmx{{ arkcase_mem_max | default('8g') }} -Djava.net.preferIPv4Stack=true -Djavax.net.ssl.trustStorePassword={{ java_trust_store_pass }} -Djavax.net.ssl.trustStore={{ java_trust_store }} -Dlog4j.formatMsgNoLookups=true"
```

Just like all the other config parameters.  Let's not get in the habit of referring to customers directly in the roles.

martinbozinovski: Thanks @david-oc-miller .

martinbozinovski: 8G should be default for ArkCase for MAX_MEM? Please advice.

david-oc-miller: No no, the default should be the old one.  8G should be the value in the Victoria PD facts file.  That way, everyone gets the old default value, except for customers like Victoria PD that need a different value.

vagrant/provisioning/arkcase-ce-facts.yml
```diff
@@ -613,4 +613,8 @@ arkcase_portal_extension: core
 arkcase_portal_extension_formatted: "{{ '-' ~ arkcase_portal_extension if arkcase_portal_extension != '' else '' }}"
 deploy_arkcase_portal: false
 
-open_office_version: "4.1.11"
\ No newline at end of file
+open_office_version: "4.1.11"
+
+# ArkCase memory size
+arkcase_mem_max: ""
+arkcase_mem_min: ""
```
dzdimov: We don't really need these facts here.

dzdimov (dismissed):

vagrant/provisioning/roles/tomcat/tasks/main.yml
martinbozinovski: Initial and max memory size is not set by us. I looked at several environments (core.dev, foia.dev. core.test) and the initial memory is between 1g(test) and 3g(dev) and max size is between 3g(test) and 8g(dev). This is close to the default values set by this pull request. Opinions?

vagrant/provisioning/arkcase-ce-facts.yml
martinbozinovski: I will remove it.

bojmil-armedia: It's OK to have them here. It serves as documentation. Just set them do the default size.

dzdimov (approved):

PR #413, opened 2022-03-31
Control center data dir

bojmil-armedia:
ACS-555 - Confluent Control Center should use data directory

### DIFF
--- a/vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
+++ b/vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
@@ -153,6 +153,14 @@
     - "confluent.metrics.topic.replication=1"
   when: not control_center_production_properties_contents.stdout is match('^' + item)
 
+- name: change control center data directory
+  become: yes
+  replace:
+    path: "{{ root_folder }}/app/control-center/control-center-production.properties"
+    regexp: "confluent.controlcenter.data.dir=/var/lib/confluent/control-center"
+    replace: "confluent.controlcenter.data.dir={{ root_folder }}/data/control-center"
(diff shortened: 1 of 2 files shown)


vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
```diff
@@ -153,6 +153,14 @@
     - "confluent.metrics.topic.replication=1"
   when: not control_center_production_properties_contents.stdout is match('^' + item)
 
+- name: change control center data directory
+  become: yes
+  replace:
+    path: "{{ root_folder }}/app/control-center/control-center-production.properties"
+    regexp: "confluent.controlcenter.data.dir=/var/lib/confluent/control-center"
+    replace: "confluent.controlcenter.data.dir={{ root_folder }}/data/control-center"
+  when: not control_center_production_properties_contents.stdout is match('/var/lib/confluent/control-center')
```
david-oc-miller: The `when` looks inverted to me; you run this task only when /var/lib/confluent/control-center is already not in the file.  Either way it's not needed; the regexp task is idempotent and won't make a change if there's no match.

david-oc-miller:
Can you link the Jira ticket in the MR description, or create a Jira ticket if you don't already have one?  I asked Naveen to always link his MRs to tickets so we have a history of why we made each change.

vagrant/provisioning/roles/confluent-platform-install/tasks/main.yml
bojmil-armedia: Yes. The when condition is not needed.

martinbozinovski (approved):

david-oc-miller (approved):

PR #425, opened 2022-05-03
ACS-560 Restart pentaho service is status 404

martinbozinovski:


vagrant/provisioning/roles/pentaho-configuration/tasks/main.yml
```diff
@@ -286,6 +286,21 @@
...
+        url: https://localhost:2002/pentaho
+        validate_certs: false
+      register: pentaho_status
+      ignore_errors: yes
+    - name: restart pentaho if status code 404
+      become: yes
+      systemd:
+        name: pentaho
+        state: restarted
+      when: pentaho_status.status == 404
```
bojmil-armedia: Should we again wait for Pentaho port to be available after restarting?

martinbozinovski: Don't have a strict answer for this. Not really, because it has plenty retries on the next tasks and tried it against several playbooks(dev, ce, all) where waiting was not needed, and yes, we should wait again for Pentaho port to be available just for pure convenience.

vagrant/provisioning/roles/pentaho-configuration/tasks/main.yml
```diff
@@ -286,6 +286,21 @@
     delay: 5
     timeout: 300
```
dzdimov: @martinbozinovski I'm guessing the playbook will never reach this point as it will probably fail in the previous task. We should put a handler in the task "wait for pentaho webapp to really be available" if it fails not to fail the whole playbook but continue on the next task that you have created.

dzdimov (approved):

PR #447, opened 2022-06-23
DSO-582 Add userAuthenticationAttributeName as default variable inste…

martinbozinovski:
…ad of hardcoded


vagrant/provisioning/roles/arkcase-ldap-config/templates/ldap.yaml
```diff
@@ -34,7 +34,7 @@
       ldapAddUserConfig:
         homeDirectory: homeDirectory
         mail: mail
-        sAMAccountName: sAMAccountName
+        {{ user_attribute_name | default(sAMAccountName) }}: {{ user_attribute_name | default(sAMAccountName) }}
```
martinbozinovski, line 37: @dzdimov Can we hardcode this key, or we need dynamic?

dzdimov (approved):

PR #452, opened 2022-07-22
DSO-606 Set basic as default with leverage for custom for projects

martinbozinovski:


vagrant/provisioning/roles/foia/templates/arkcase-portal-server.yaml
```diff
@@ -29,7 +29,7 @@ portal.truststoreFilename: arkcase.ts
 portal.truststorePassword: "{{ java_trust_store_pass }}"
 
 # authentication type, possible values (basic, external)
-portal.authenticationType: basic
+portal.authenticationType: "{{ authentication_type | default('basic') }}"
```
david-oc-miller, line 32: Can the fact be named `portal_authentication_type`, or even `foia_portal_authentication_type`?  Plain `authentication_type` could refer to almost any service.  Not obvious it is for the FOIA portal.  Or even `foia_portal_to_arkcase_authentication_type`.

martinbozinovski: Yes, portal_authentication_type is better naming. Thanks!

david-oc-miller (approved):

PR #454, opened 2022-07-28
DSO-613 Installer support for the OIDC microservice

david-oc-miller:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -911,6 +911,15 @@
     dest: /home/arkcase/.arkcase/acm/acm-config-server-repo/arkcase-commons-server.yaml
   when: arkcase_version == "" or arkcase_version is version('2021.04', '>=')
 
+- name: server-specific configuration for arkcase-portal-oidc.yml for version 2021.04
```
martinbozinovski, line 914: As i am aware of, ArkCase is at 2022.01-RC1. Are we going with 2021.04?

martinbozinovski (commented):
Other that arkcase_version, being 2021.04, this looks good!.

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
david-oc-miller: 2021.04 is the convention used in other version checks for any version higher than 2021.03, including 2022 and future versions.

martinbozinovski: Perfect than. Thanks for the clarification!.

martinbozinovski (approved):

mariogjurcheski (approved):

PR #459, opened 2022-08-04
DSO-609 Add portal OU's and groups

dzdimov:


david-oc-miller:
This code looks OK.  I believe we also need changes to the arkcase-app role, to include the core portal directory configuration in the ldap-server.yaml.  This isn't being done now, so for example on PRC dev, the default configuration (which refers to ldaps://acm-arkcase:636) is still being used, so on every LDAP sync we see "acm-arkcase host name not found" exceptions.

david-oc-miller (approved):

PR #460, opened 2022-08-08
changed when condition

dzdimov:


martinbozinovski (approved):

vagrant/provisioning/roles/ldap/tasks/main.yml
```diff
@@ -101,7 +101,7 @@
   loop:
     - "{{ ldap_core_portal_group_no_base }},{{ ldap_base }}"
     - "{{ ldap_core_portal_user_no_base }},{{ ldap_base }}"
-  when: enable_microservices is true
+  when: enable_microservices is defined
```
david-oc-miller, line 104: what happens when `enable_microservices` is defined, and the value is `false`?

PR #463, opened 2022-08-12
Libre Office install

sparksam:
FileUtils and FileManagement url paths updates and officehome path update.

### DIFF
--- a/vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
+++ b/vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
@@ -21,4 +21,4 @@ arkcase:
 jodconverter:
   local:
     enabled: true
-    officeHome: /opt/libreoffice/program
+    officeHome: /opt/libreoffice7.3/
--- a/vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
+++ b/vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
@@ -1,7 +1,8 @@
 arkcase:
   baseURL: "https://{{ external_host }}/arkcase"
   basePortalURL: "https://{{ external_host }}/portal"
-  fileManagementURL: "https://{{ external_host }}"
+  fileManagementURL: "https://{{ internal_host }}:{{ file_management_port | default(9068) }}"
+  fileUtilsURL: "https://{{ internal_host }}:{{ file_utils_port | default(9078) }}"
   enableKafkaSync:
     schema: {{ enableSchemaSync | default(false) }}
     caseFile: {{ enableCaseFileSync | default(false) }}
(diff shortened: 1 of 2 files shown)


vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -75,24 +75,24 @@
       dest: "/tmp/LibreOffice_{{ libreoffice_version | default('7.3.4') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
       owner: root
       url: "https://ftp.osuosl.org/pub/tdf/libreoffice/stable/{{ libreoffice_version | default('7.3.4') }}/rpm/x86_64/LibreOffice_{{ libreoffice_version | default('7.3.4') }}_Linux_x86-64_rpm.tar.gz --no-check-certificate"
-  when: not libreoffice_bin_exists
+  when: libreoffice_bin_exists
```
david-oc-miller: What is your purpose with this change?  I'm not sure how the code worked, it seems the check should be more like: `when: not libreoffice_bin_exists.stat.exists`, see https://docs.ansible.com/ansible/latest/collections/ansible/builtin/stat_module.html#return-values

david-oc-miller: and same comment for all the other changes.

vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
```diff
@@ -1,18 +1,24 @@
 arkcase:
   baseURL: "https://{{ external_host }}/arkcase"
   basePortalURL: "https://{{ external_host }}/portal"
-  fileManagementURL: "https://{{ external_host }}"
+  fileManagementURL: "https://{{ internal_host }}:{{ file_management_port | default(9068) }}"
+  fileUtilsURL: "https://{{ internal_host }}:{{ file_utils_port | default(9078) }}"
```
david-oc-miller, line 5: was this called fileUtilsMicroserviceURL before?  What value is the arkcase-file-utils service looking for?

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
sparksam: I removed the when because I found out Ansible automatically skips the tasks when Libreoffice is already installed.

vagrant/provisioning/roles/arkcase-app/templates/arkcase-commons-server.yaml
sparksam: Yes. I updated the field name and updated in the prc extension.

vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -62,3 +62,36 @@
...
+      dest: "/tmp/LibreOffice_{{ libreoffice_version | default('7.3.4') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
+      owner: root
+      url: "https://ftp.osuosl.org/pub/tdf/libreoffice/stable/{{ libreoffice_version | default('7.3.4') }}/rpm/x86_64/LibreOffice_{{ libreoffice_version | default('7.3.4') }}_Linux_x86-64_rpm.tar.gz --no-check-certificate"
+
+- name: unpack LibreOffice RPM
+  become: yes
+  unarchive:
+    src: "/tmp/LibreOffice_{{ libreoffice_version | default('7.3.4') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
+    remote_src: yes
+    dest: /tmp/
```
david-oc-miller, line 85: These commands will run every time, which will be slow since LibreOffice is such a large package.  You could add a `when: not libreoffice_bin_exists.stat.exists`

sparksam: Thanks Dave. Just did.

david-oc-miller (approved):

PR #478, opened 2022-08-29
DSO-658: adjustments for Cognito microservice changes

david-oc-miller:


martinbozinovski:
What about left overs, services, yaml files? Do we need cleanup on that? Otherwise, looks good!

david-oc-miller:
> What about left overs, services, yaml files? Do we need cleanup on that? Otherwise, looks good!

Good catch, thanks.  New commit pushed.

martinbozinovski (approved):

PR #501, opened 2022-10-03
Few fixes

sparksam:
- Backup of server config should not fail the deployment.


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -902,6 +902,7 @@
     backup: yes
     src: server.yml
     dest: /home/arkcase/.arkcase/acm/acm-config-server-repo/arkcase-server.yaml
+  ignore_errors: true
```
david-oc-miller, line 905: What kind of errors are we ignoring here?  Generally I prefer a specific `failed_when` clause, to specify the exact errors that should not cause the task to fail.

PR #506, opened 2022-10-20
DSO-734 Add when clauses if db_engine is postgresql

martinbozinovski:


vagrant/provisioning/arkcase-dev-facts.yml
```diff
@@ -717,6 +717,7 @@ postgres_log_folder: "{{ root_folder }}/log/postgresql"
 postgresql_auth_method: "{{ ansible_fips  | ternary('scram-sha-256', 'md5') }}"
 postgresql_user: postgres
 postgresql_group: postgres
+postgresql_unix_socket_folder: /var/run/postgresql
```
david-oc-miller, line 720: This will work only when PostgreSQL is installed locally.  It won't work when PostgreSQL is on a different host.  There should be some authentication method that will work whether PostgreSQL is local or remote.  If there isn't, then we need two separate blocks (one for local using the socket, another for remote using TCP), with a flag to control which block is used.

dzdimov (approved):

PR #516, opened 2022-12-07
DSO-768 Add variables for pdfTron license

martinbozinovski:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -1154,4 +1154,11 @@
 - name: configure Prc-Dockets, if needed
   include_tasks: configure_prc_dockets.yml
 
-
+- name: add pdfTron license if defined
+  become: yes
+  become_user: arkcase
+  lineinfile:
```
david-oc-miller, line 1172: This will update the file every time, since the line includes the regexp. It is not idempotent, which we generally shoot for.  But at least it doesn't add onto the line... the file will be updated with the same text every time.

vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -159,3 +162,8 @@ payment:
     securepaylinkendpoint: "{{ touchnet_securepaylinkendpoint | default('') }}"
     upaysiteid: "{{ touchnet_upaysiteid | default('') }}"
 microsoftOfficeOnline.isWopiEnabled: {{ arkcase_ee | default('false') }}
+
+#### Configuration variables for PDFTron integration ###########
+pdftron:
+  viewer:
+    license: "{{ pdftron_viewer_license | default('') }}"
```
david-oc-miller, line 169: Why do we need `pdftron_viewer_license` in two places?

martinbozinovski: Don't really know, it was requirements from Nadica, those lines were in .arkcase MR's that she pushed.

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
martinbozinovski: Dave, thanks for the comment. I update the code accordingly.

david-oc-miller (approved):

PR #530, opened 2023-01-17
Additional properties for ZyLab new authentication flow.

venkatarmedia:


vagrant/provisioning/roles/oidc-config/templates/arkcase-oidc-server.yaml
```diff
@@ -13,3 +13,5 @@ oidc:
   systemUserEmail: "{{ oidc_systemUserEmail }}"
   systemUserPassword: "ENC({{ encrypted_oidc_systemUserPassword }})"
   usersDirectory: "{{ ldap_user_domain | replace('.', '_') }}"
+  responseType: "{{ oidc_responseType }}"
```
david-oc-miller: @venkatarmedia sorry I didn't see this MR until now.  What would the values be for these two new variables?  Can you supply meaningful default values here, so sites whose config files don't have `oidc_responseType` and `oidc_responseMode`, won't break?

venkatarmedia: @david-oc-miller, added default values.

dzdimov (approved):

david-oc-miller (approved):

PR #538, opened 2023-02-02
Remove the new ignore_errors attributes

david-oc-miller:
Revert the recent ignore_errors additions (totally different semantics than 'warn: false').

In future, do not merge installer changes until I review them... I'll change the github merge rules if I have to.


martinbozinovski (commented):
So, we will be left without warns and will have fails? That is the desired outcome?

david-oc-miller:
@martinbozinovski you are not understanding what `warn: false` does.  All it does, is prevent Ansible saying, "You should use get_url instead of curl".  That is it.  ignore_errors is much broader.  We don't want to ignore errors... if we get an error, the play should fail, since it did fail.  Due to the error.  "on error resume next" is an antipattern.

david-oc-miller:
Please do not make changes without a correct understanding of what you're actually doing.

martinbozinovski (approved):

PR #546, opened 2023-03-03
Add libraries in idempotent way

martinbozinovski:


dzdimov (dismissed):

vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -1190,6 +1190,13 @@
   environment:
     SSHPASS: "{{ sftp_arkcase_password }}"
 
+- name: create libraries directory if it does not exist
+  become: yes
+  become_user: arkcase
+  file:
+    path: "{{ root_folder }}/data/arkcase-home/.arkcase/libraries"
```
david-oc-miller: In this file, all other references to the .arkcase folder use `/home/arkcase/.arkcase`, which would be either a regular folder or a symlink to  `"{{ root_folder }}/data/arkcase-home/.arkcase`.  Let's use `/home/arkcase/.arkcase` here too, for consistency.

david-oc-miller (approved):

PR #549, opened 2023-03-09
DSO-783 Installer will use alfresco 7.3 ee

martinbozinovski:


vagrant/provisioning/roles/alfresco7/tasks/apply-amp.yml
```diff
@@ -14,4 +14,23 @@
       become_user: alfresco7
       command: java -jar "{{ root_folder }}/app/alfresco7/bin/alfresco-mmt.jar" install {{ amp.amp_path }} "{{ root_folder }}/app/alfresco7/webapps/{{ amp.target_war }}"
       when: 'amp.title not in installed_amps.stdout'
+  when: alfresco_ee_version is version('7.3.0.1', '<')
```
martinbozinovski, line 17: @david-oc-miller @dzdimov Do we need the option for older alfresco versions? 7.3 is working with java-11, and if we are going with alfresco7.3 from now on, i can just leave one set of tasks here, only for java-11, and delete the ones with java8

david-oc-miller: I don't know; we would have to try it and see.  When we want to run the installer to fix some issue with existing, older Alfresco installations, but we don't want to upgrade to the newer Alfresco, then it makes sense to continue supporting older Alfresco versions.

vagrant/provisioning/roles/alfresco7/tasks/main.yml
```diff
@@ -274,7 +274,7 @@
       messaging.broker.username={{ activemq_user | default('guest', true) }}
       messaging.broker.password={{ activemq_password | default('guest', true) }}
       index.subsystem.name=solr6
-      solr.secureComms=none
+      solr.secureComms=https
```
david-oc-miller, line 277: Does Alfresco-to-Solr communication work with this change?  More changes may be needed to enable TLS between Alfresco Content Services and Alfresco Search Services.

david-oc-miller:
Do these changes work for both Alfresco CE and Alfresco EE?

vagrant/provisioning/roles/alfresco7/tasks/main.yml
martinbozinovski: For local arkcase installment, yes. That is why i put this PR in WIP, so this kind of thoughts be shared in order to make adjustment to the code.

martinbozinovski:
I've tried only with EE :(. I will test out with CE.

martinbozinovski:
@david-oc-miller These changes are working for both EE and CE.

david-oc-miller (approved):

PR #552, opened 2023-04-04
Installer changes for alfresco upgrage

aleksandar-bujaroski-armedia:


martinbozinovski:
@aleksandar-bujaroski-armedia This is combination of manual(deletion files, folders) and automation (mariadb in alfresco role) work, as i am aware of. It will be good practice to try on some test server, even dev will be fine before going to prod, pre-prod.

aleksandar-bujaroski-armedia:
> @aleksandar-bujaroski-armedia This is combination of manual(deletion files, folders) and automation (mariadb in alfresco role) work, as i am aware of. It will be good practice to try on some test server, even dev will be fine before going to prod, pre-prod.

@martinbozinovski yes, this first needs to be tried on DEV,TEST before going to live environments

PR #554, opened 2023-04-05
DSO-879 Make changes for support of Rocky8

martinbozinovski:
@david-oc-miller Not sure if to add "or" in when checks or a separate task. 
Also, i need to test it on centos7 with this branch and using the ansible_python_interpreter: /usr/bin/python3. That will mainly depends on ansible version, since on Rocky8 python version for ansible is 3.9. 
That is why this is DRAFT.

Edit: python version = 2.7.5 on centOS7. So, new tasks if we continue supporting centOS7

### DIFF
--- a/vagrant/Vagrantfile.rocky8
+++ b/vagrant/Vagrantfile.rocky8
@@ -3,30 +3,6 @@
 
 # Vagrant.require_version ">= 2.1.0" # 2.1.0 minimum required for triggers
 
-# user = ENV['RH_SUBSCRIPTION_MANAGER_USER']
-# password = ENV['RH_SUBSCRIPTION_MANAGER_PW']
-
-# register_script = %{
-# if ! sudo subscription-manager status; then
-#   if [ "#{user}" = "" ] || [ "#{password}" = "" ]; then
-#     echo 'Required environment variables not found. Please set RH_SUBSCRIPTION_MANAGER_USER and RH_SUBSCRIPTION_MANAGER_PW'
-#     exit 1
-#   fi
-#   echo "registering..."
-#   sudo subscription-manager register --username=#{user} --password='#{password}' --auto-attach 
-# fi
-# }
-
-# unregister_script = %{
-#   if [ "#{user}" = "" ] || [ "#{password}" = "" ]; then
-#     echo 'Required environment variables not found. Please set RH_SUBSCRIPTION_MANAGER_USER and RH_SUBSCRIPTION_MANAGER_PW'
-#     exit 1
-#   fi
-# if sudo subscription-manager status; then
-#   sudo subscription-manager unregister
-# fi
-# }
-
 # All Vagrant configuration is done below. The "2" in Vagrant.configure
 # configures the configuration version (we support older styles for
 # backwards compatibility). Please don't change it unless you know what
--- a/vagrant/Vagrantfile.rocky8
+++ b/vagrant/Vagrantfile.rocky8
@@ -1,0 +1,153 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.require_version ">= 2.1.0" # 2.1.0 minimum required for triggers
+
+user = ENV['RH_SUBSCRIPTION_MANAGER_USER']
+password = ENV['RH_SUBSCRIPTION_MANAGER_PW']
+
+register_script = %{
+if ! sudo subscription-manager status; then
+  if [ "#{user}" = "" ] || [ "#{password}" = "" ]; then
+    echo 'Required environment variables not found. Please set RH_SUBSCRIPTION_MANAGER_USER and RH_SUBSCRIPTION_MANAGER_PW'
+    exit 1
+  fi
+  echo "registering..."
+  sudo subscription-manager register --username=#{user} --password='#{password}' --auto-attach 
+fi
+}
+
+unregister_script = %{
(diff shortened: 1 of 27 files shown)


vagrant/provisioning/roles/alfresco-site/tasks/main.yml
```diff
@@ -70,14 +70,18 @@
     password: "{{ alfresco_admin_password }}"
   register: rm_container_out
 
+- name: install jmespath for python3.9
+  become: yes
+  command: python3.9 -m pip install jmespath
+
```
dzdimov, line 82: Do we need for all OS's or just Rocky?

david-oc-miller: jmespath we need on all systems, the Alfresco site and folder role uses it.  I have to install it manually on new sites, before I can finish the alfresco role.

martinbozinovski: Yes. This is fix on the next commit, i changed this to have when clause. We just need to install it when ansible is invoked by python3.9.

vagrant/Vagrantfile.rocky8
```diff
@@ -0,0 +1,153 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.require_version ">= 2.1.0" # 2.1.0 minimum required for triggers
+
+user = ENV['RH_SUBSCRIPTION_MANAGER_USER']
+password = ENV['RH_SUBSCRIPTION_MANAGER_PW']
+
+register_script = %{
```
david-oc-miller: For Rocky 8 we shouldn't have to register with Red Hat Network.

vagrant/provisioning/roles/mariadb/tasks/main.yml
```diff
@@ -29,6 +29,22 @@
...
+    content: |
+      # MariaDB 10.4 RedHat repository list - created  2022-01-07
+      # https://mariadb.org/download/
+      [mariadb]
+      name = MariaDB
+      baseurl = http://yum.mariadb.org/10.6/rhel8-amd64
+      module_hotfixes=1
+      gpgkey=https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
+      gpgcheck=1
+  when: ansible_distribution == "Rocky" and ansible_distribution_major_version == "8"
```
david-oc-miller, line 43: Should we also check for RedHat here?

vagrant/provisioning/roles/samba/tasks/main.yml
```diff
@@ -77,20 +88,20 @@
       - gcc 
       - gdb 
       - gnutls-devel 
-      - gpgme-devel 
+      - gpgme-devel #failed
```
david-oc-miller: These packages shouldn't fail any more, after enabling powertools, right?

david-oc-miller:
> Not sure if to add "or" in when checks or a separate task.

The "or" checks look OK to me.

Can we set some kind of global fact with the desired python executable (depending on which OS it is), and use that fact everywhere?

vagrant/provisioning/roles/mariadb/tasks/main.yml
martinbozinovski: Well, the baseurl is different than the above for RedHat and also the gpgkey. And i didn't want to change the one for Redhat, just add one that is Rocky specific.

martinbozinovski:
Yes Dave. That is something that i refactor on currently and will provide better support for different  OS. Since centOS ansible python is 2.7 and Rocky is 3.9, that should be configurable depending of the OS.

vagrant/provisioning/roles/samba/tasks/main.yml
martinbozinovski: I removed this comment in the upcoming commit, i forgot them in this one. And yes, they don't fail anymore.

vagrant/provisioning/roles/mariadb/tasks/main.yml
david-oc-miller: That makes sense, thanks.  Can you fix the task name, so it mentions Rocky, versus RHEL 8?

david-oc-miller:
@martinbozinovski please review latest changes.  This gets the BCGEU ArkCase installed on my local Rocky 8 VM.

martinbozinovski:
> @martinbozinovski please review latest changes. This gets the BCGEU ArkCase installed on my local Rocky 8 VM.

Sure Dave. Do i have deadline on this? It would be nice if i have time till the end of the week, to test it properly.
I saw it in commit bad1ca75234ced42044e9765356a407945bb9532 that MariaDB support will be to 8.8 or higher. 
We will be supporting BCGEU with that version/s for all of the packages?

david-oc-miller:
> > @martinbozinovski please review latest changes. This gets the BCGEU ArkCase installed on my local Rocky 8 VM.
> 
> Sure Dave. Do i have deadline on this? It would be nice if i have time till the end of the week, to test it properly. I saw it in commit [bad1ca7](https://github.com/ArkCase/arkcase-ce/commit/bad1ca75234ced42044e9765356a407945bb9532) that MariaDB support will be to 8.8 or higher. We will be supporting BCGEU with that version/s for all of the packages?

I'll let BCGEU know the Rocky 8 support is in test, and that we require 8.8 or higher.

david-oc-miller (approved):

PR #555, opened 2023-04-13
DSO-885 Add check for file presence

martinbozinovski:


vagrant/provisioning/roles/arkcase-app/tasks/main.yml
```diff
@@ -744,7 +751,7 @@
       replace: "ENC({{ encrypted_default_database }})"
     - regexp: "ENC.U2FsdGVkX1./i7RReaawm2OyEcse1DBxZ0.oGE3ZsT5V1ahjvWhorhAZoUB97sHH."
       replace: "ENC({{ encrypted_snowbound_key }})"
-  when: arkcase_version != "" and arkcase_version is version('2021.04', '<')
+  when: forms_file_check.stat.exists == true and arkcase_version != "" and arkcase_version is version('2021.04', '<')
```
david-oc-miller, line 754: This is fine as it is, but you don't need the `== true` part.  You could just say: `forms_file_check.stat.exists and arkcase_version != "" and arkcase_version is version('2021.04', '<')`

david-oc-miller (approved):

PR #562, opened 2023-07-17
ACS-1332 Add List support for ldapUrls

martinbozinovski:


vagrant/provisioning/arkcase-ce-facts.yml
```diff
@@ -50,7 +50,8 @@ default_user_password: "@rKc@3e"
 ldap_host: "{{ internal_host }}"
 ldap_protocol: ldaps
 ldap_port: 636
-ldap_url: "{{ ldap_protocol }}://{{ ldap_host }}:{{ ldap_port }}"
```
david-oc-miller, line 53: Do we need version detection here?  The ArkCase change seems like only in `develop` branch, which is the blank version and version 2023.02 and above.  Versions 2023.01 and below still need the old scalar non-list variable right?

martinbozinovski: They've import the changes into 2023.01.xx. I can see the ldapUrl is List in ldap.yaml. Probably we will need support for ArkCase-CE, that is why i pushed the changes here.

david-oc-miller: Yes, but we still have customers on 2021.02 and 2021.03.  The installer still needs to support these older versions.  I guess these 3 facts files are not used for the older installations... Maybe as long as the fact file matches the target configuration it will work either way, so it might be OK as is.

martinbozinovski: Nice point. So, if new client takes the arkcase-ce, will need List, and if client wants to run ldap again, will need String. Got it, thanks Dave for point that. I will leave it in Draft for current day. Or, maybe this to continue, and just to remember facts file for older installation to be String for ldapUrl. Because, i've changed only facts file, not roles/playbooks.

PR #563, opened 2023-07-20
ACS-1344 Change the support for Linux OS

martinbozinovski:


vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
```diff
@@ -136,7 +136,7 @@
 - name: setup path to arkcase config file
   become: yes
   become_user: pentaho
-  shell: sed -i 's~C:\\Users\\marjan\.stefanoski\\foia---reports-dw-1\\config\\arkcase_config\.properties~{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/config/arkcase_config\.properties~g' {{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/misc/VARIABLE_CONFIG.ktr
+  shell: sed -i 's~/home/marst/projects/ETL-projects/EEOC/foia---reports-dw-1/config/arkcase_config\.properties~{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/config/arkcase_config\.properties~g' {{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/misc/VARIABLE_CONFIG.ktr
```
david-oc-miller, line 139: I guess Marjan switched to a Linux laptop.  Do we need a version check?  Older foia-reports-dw versions will still have the Windows paths.  Or else we could run both `sed` commands (the Windows one and the Linux one).

marjan4jc: @david-oc-miller , all versions are using the Linux path.

david-oc-miller: Hmm... well, we can always restore the Windows path changes if needed.

david-oc-miller (dismissed):

vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
marjan4jc: There will be no need for that, PDI 9.4, which we are using in k8S is supporting VARIABLES in these 3 types of transformations where we use "sed" now, so with k8s, the issue is solved by upgrading. If we upgrade pentaho and PDI to 9.4 on other environments, we can remove sed commands in the installer as well.

marjan4jc (approved):

vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
david-oc-miller: @marjan4jc , @martinbozinovski , I still think we need to run both the Windows `sed` and the Linux `sed` commands, to replace both kinds of paths; or else we need to see whether the file contains Windows paths or Linux paths, and then run the appropriate replacement.  Some client may have an older version of the reports with the old Windows paths.

martinbozinovski: @david-oc-miller, you are correct on this. This is called with arkcase-app, which is part of the upgrade strategy, and will be called every time when we upgrade and environment. I will make changes and place a when condition with something like: reports_version or arkcase_version. Will talk to @marjan4jc and place this ticket in Draft/WIP for now.

martinbozinovski: @david-oc-miller i don't have clue why pentaho stuff got committed here. My idea is to just add the "linux" tasks and keep the windows, because they are shell commands, and such as, if no path is found, will not stop ansible from going further, vs. ansible tasks which will stop the installation if they fail (except ignore if failing).

david-oc-miller: @martinbozinovski add the linux tasks and keep the windows ones, exactly.  Sounds good.  So whichever format the paths are in, they'll get fixed appropriately.

PR #573, opened 2023-09-05
MariaDB RHEL 8.8 fixes

david-oc-miller:


david-oc-miller:
I'm going to merge this now, without outside review, since I need it for JCT tomorrow and because I tested it locally.

PR #574, opened 2023-09-06
Config server: use ActiveMQ config for 2023.01 and later

david-oc-miller:


david-oc-miller:
I don't remember whether new ArkCase uses Kafka or AMQ, but then, new ArkCase isn't using Ansible apparently. 

Merging this now due to JCT needs.

PR #580, opened 2023-10-06
Fixed library path for pdfTron libraries

venkatarmedia:


martinbozinovski (approved):

vagrant/provisioning/roles/arkcase-prerequisites/tasks/main.yml
```diff
@@ -212,6 +212,6 @@
     insertbefore: EOF
     marker: "# {mark} ANSIBLE MANAGED BLOCK - PdfTron"
     block: |
-      export JAVA_OPTS="$JAVA_OPTS -Djava.library.path={{ root_folder }}/data/arkcase-home/.arkcase/libraries:{{ root_folder }}/app/arkcase/lib"
```
david-oc-miller, line 215: no more need to root/app/arkcase/lib?

venkatarmedia: yes, we have a lot of stuff here which we don't need. We have everything in the libraries folder.

venkatarmedia: Even if we require something its better to keep it under the libraries folder instead of having it under root/app/arkcase/lib.

PR #581, opened 2023-10-26
remove EEOC from the path

martinbozinovski:


vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
```diff
@@ -141,7 +141,7 @@
 - name: setup path to arkcase config file(Linux)
   become: yes
   become_user: pentaho
-  shell: sed -i 's~/home/marst/projects/ETL-projects/EEOC/foia---reports-dw-1/config/arkcase_config\.properties~{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/config/arkcase_config\.properties~g' {{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/misc/VARIABLE_CONFIG.ktr
```
david-oc-miller, line 144: On all these things, let's add an extra step, versus removing one, to maintain maximum compatibility with all report versions.  Like, here we could add a loop, and run two `sed` commands, one with EEOC and one without it.

martinbozinovski: Or, just ansible lineinfile to replace whatever expression we give it, like "/home/marst/*" ?? That way, we will handle anything that is in the file, rather than separate EEOC and whatnot. And, it is ansible module, so...Give it a thought Dave, and i will change the code accordingly.

david-oc-miller: 👍 lineinfile, excellent idea.

vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
```diff
@@ -136,12 +136,22 @@
...
+    regexp: '^(\s*)<name>C:\\Users\\marjan\.stefanoski\\foia---reports-dw-1\\config\\arkcase_config.properties</name>'
+    line: '\1<name>{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/config/arkcase_config.properties</name>'
 
 - name: setup path to arkcase config file(Linux)
   become: yes
   become_user: pentaho
-  shell: sed -i 's~/home/marst/projects/ETL-projects/EEOC/foia---reports-dw-1/config/arkcase_config\.properties~{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/config/arkcase_config\.properties~g' {{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/misc/VARIABLE_CONFIG.ktr
+  lineinfile:
+    path: "{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}/misc/VARIABLE_CONFIG.ktr"
+    state: present
```
david-oc-miller: Do these new regexps still work when `EEEOC` is in the line?

martinbozinovski: No Dave, they did not work work for anything else beside that :(. I refactor and provide commit for including more wide search, and since "marst" is pretty specific term, we should be ok. I did couple of tests locally, it went ok.

david-oc-miller: One last question ... does `*` work as expected, or do you need `.*`?

martinbozinovski: I think that i hit the spot with the last commit. But, needed rearrange because the task "replace developer local path with good local path" was interfering on the task "setup path to arkcase config file(Linux)"

david-oc-miller (approved):

PR #583, opened 2023-11-02
Initial commit for solr upgrade

martinbozinovski:
We can add delete of solr_backup at last, just for cleanup (i would leave the solr_backup because of not covered scenarios, just for backup and do a manual remove if everything is successful).
@david-oc-miller  any thoughts on this whole PR?
Also, i would want to test it on test core/foia envs, since they will be on k8s next week, so we can play with "real data" on this upgrade.

### DIFF
--- a/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
@@ -1,65 +1,91 @@
-- name: stop arkcase
+- name: check for available disk space
   become: yes
-  systemd:
-    daemon_reload: true
-    name: arkcase
-    state: stopped
+  command: df -h /opt/
+  register: space_available
 
-- name: copy existing solr data to backup folder
+- name: check for solr data space
   become: yes
-  copy:
-    remote_src: yes
-    src: "{{ root_folder }}/data/solr/"
-    dest: "{{ root_folder }}/data/solr_backup/"
-    owner: solr
-    group: solr
+  stat:
+    path:  /opt/arkcase/data/solr
+  register: solr_data
 
-- name: delete solr cores from previous install
-  command: /opt/arkcase/app/solr/bin/solr delete -c {{ item }}
-  loop: "{{ solr_cores }}"
+- name: upgrade and backup solr if space is available under root_folder
+  block: 
+    - name: stop arkcase
+      become: yes
+      systemd:
+        name: arkcase
+        state: stopped
 
-- name: stop solr
+    - name: copy existing solr data to backup folder
+      become: yes
+      copy:
+        remote_src: yes
+        src: "{{ root_folder }}/data/solr/"
+        dest: "{{ root_folder }}/data/solr_backup/"
+        owner: solr
+        group: solr
+
+    - name: delete solr cores from previous install
+      command: /opt/arkcase/app/solr/bin/solr delete -c {{ item }}
+      loop: "{{ solr_cores }}"
+
+    - name: stop solr
+      become: yes
+      systemd:
+        name: solr
+        state: stopped
+
+    - name: Remove solr config files and folders
+      become: yes
+      file:
+        path: "{{ item }}"
+        state: absent
+        group: solr
+        owner: solr
+      loop:
+        - "{{ root_folder }}/data/solr"
+        - "{{ root_folder }}/app/solr"
+        - "{{ root_folder }}/install/solr"
+        - "/etc/systemd/system/solr.service"
+
+    - name: install solr
+      include_role: 
+        name: solr
+
+    - name: remove solr default data folder after upgrade
+      become: yes 
+      shell: rm -rf {{ root_folder }}/data/solr/*
+
+    - name: copy backup dir to new solr data dir
+      become: yes
+      copy:
+        remote_src: yes
+        src: "{{ root_folder }}/data/solr_backup/"
+        dest: "{{ root_folder }}/data/solr/"
+        owner: solr
+        group: solr
+
+    - name: restart service solr
+      become: yes
+      systemd:
+        daemon_reload: true
+        name: solr
+        state: restarted
+      
+    - name: wait for solr to start before start arkcase service
+      pause:
+        minutes: 1
+
+    - name: start service arkcase
+      become: yes
+      systemd:
+        name: arkcase
+        state: started
+  when: solr_data < space_available
+
+- name: output if there is not space available
   become: yes
-  systemd:
-    daemon_reload: true
-    name: solr
-    state: stopped
-
-- name: Remove solr config files and folders
-  become: yes
-  file:
-    path: "{{ item }}"
-    state: absent
-    group: solr
-    owner: solr
-  loop:
-    - "{{ root_folder }}/data/solr"
-    - "{{ root_folder }}/app/solr"
-    - "{{ root_folder }}/install/solr"
-    - "/etc/systemd/system/solr.service"
-
-- name: install solr
-  include_role: 
-    name: solr
-
-- name: remove solr default data folder after upgrade
-  become: yes 
-  shell: rm -rf {{ root_folder }}/data/solr/*
-
-- name: copy backup dir to new solr data dir
-  become: yes
-  copy:
-    remote_src: yes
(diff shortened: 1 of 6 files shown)


vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
+- name: stop arkcase
+  become: yes
+  systemd:
+    daemon_reload: true
+    name: arkcase
+    state: stopped
+
+- name: copy existing solr data to backup folder
```
david-oc-miller: Want to check existing disk space here?  If the free space is less than the Solr data folder size, this part would fail.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: install solr
+  include_role: 
+    name: solr
+
+- name: remove solr default data folder after upgrade
+  become: yes 
+  shell: rm -rf {{ root_folder }}/data/solr/*
+
+- name: copy backup dir to new solr data dir
+  become: yes
```
david-oc-miller: Could this be a move instead of a copy?

david-oc-miller:
I integrated my tomcat upgrade code into the tomcat role itself; the tomcat role compares the current installed version to the desired version, and if they differ, it moves the current folder.... this depends on Tomcat upgrades being backward compatible.  If you can detect the Solr version, you could take a similar approach here, and then the only work needed is to update the Solr version in the facts file.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
martinbozinovski: You are right. Will modify accordingly.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
martinbozinovski: Well, it could be, but this is brand new data directory, so no backed up data here, just what do we get after installation. So, no real point of backing up on that. Or, maybe i am missing something...

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
+- name: check for available disk space
+  become: yes
+  command: df -h /opt/
```
david-oc-miller: here you want to use `{{ root_folder }}` vs `/opt`.  Sometimes `root_folder` is /opt, but more often it is `/opt/app` and on FedRAMP sites, `/opt/app` is its own file system.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+  register: space_available
+
+- name: check for solr data space
+  become: yes
+  stat:
+    path:  /opt/arkcase/data/solr
```
david-oc-miller: Same here, check `{{ root_folder }}/data/solr`

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+
+- name: upgrade and backup solr if space is available under root_folder
+  block: 
+    - name: stop arkcase
+      become: yes
+      systemd:
+        name: arkcase
+        state: stopped
+
+    - name: copy existing solr data to backup folder
```
david-oc-miller, line 24: first stop Solr, then backup the data and delete the cores, right?

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+
+    - name: remove solr default data folder after upgrade
+      become: yes 
+      shell: rm -rf {{ root_folder }}/data/solr/*
+
+    - name: copy backup dir to new solr data dir
+      become: yes
+      copy:
+        remote_src: yes
+        src: "{{ root_folder }}/data/solr_backup/"
```
david-oc-miller, line 68: The Solr backup folder maybe quite large.  With a move, all that happens is the directory name is changed.  With a copy, it will take some time to copy the data.  Also, won't this copy just create `{{ root_folder }}/data/solr/solr_backup`?

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+        group: solr
+
+    - name: restart service solr
+      become: yes
+      systemd:
+        daemon_reload: true
+        name: solr
+        state: restarted
+      
+    - name: wait for solr to start before start arkcase service
```
david-oc-miller: We have other tasks that first wait for the port to be open, and then waits for some given URL to return a 200, versus only waiting a minute.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+      pause:
+        minutes: 1
+
+    - name: start service arkcase
+      become: yes
+      systemd:
+        name: arkcase
+        state: started
+  when: solr_data < space_available
```
david-oc-miller: The structure returned by `stat` does not include the total amount of space used under a folder, and the stdout from `dh -h` is a formatted string e.g.  
  

```
Filesystem            Size  Used Avail Use% Mounted on
/dev/mapper/vg01-opt 1014M   33M  982M   4% /opt
```

This command returns only a number with the used space of a file system:   
  
`df {{ root_folder }}/data/solr  --output=used | grep -v Used`
  
While this command returns only a number with the space used under a folder:  
  
`du {[ root_folder }}/solr/data -s | awk '{ print $1 }'`
  
To compare the two command outputs, something like this:  
  
`when: solr_data.stdout | int < space_available.stdout | int`

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
martinbozinovski: No. We can't delete the cores if Solr is not running. So, backup data, delete cores, stop Solr.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
martinbozinovski: It will copy the files from backup to solr directory because of the slash in the end, and not the directories itself. 
However, next task will wait on this one, as much is needed to be finished.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -1,14 +1,18 @@
 - name: check for available disk space
   become: yes
-  command: df -h "{{ root_folder }}/"
+  shell: df -h "{{ root_folder }}/" | awk 'NR==2 { print $4 }'
```
martinbozinovski, line 3: This will print the 4th row from df -h, which is Avail row and its data

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -1,14 +1,18 @@
...
   become: yes
-  stat:
-    path:  "{{ root_folder }}/data/solr"
+  shell: du -sh "{{ root_folder }}/data/solr" | awk '{ print $1 }'
   register: solr_data
 
+- name: format free space and solr data size output
+  set_fact:
+    space_available_formated: "{{ space_available.stdout | human_to_bytes | int }}"
+    solr_data_formated: "{{ solr_data.stdout | human_to_bytes | int }}"
```
martinbozinovski, line 14: This is from debugger, how the output will be look like for when conditions below.
TASK [upgrade-solr : debug space] ************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "102005473280"
}

TASK [upgrade-solr : debug solr] *************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "716800"
}

david-oc-miller (approved):

PR #584, opened 2023-11-13
Update tomcat and custom libraries through running upgrade tag

martinbozinovski:
For this to be applicable, tomcat_version in facts file need to be changed, and the ansible installer run with "upgrade" tag. 
If something went wrong,  i am leaving backup of /app/arkcase deliberately, and can be delete after successful upgrade and confirmed working of application.

### DIFF
--- a/vagrant/provisioning/arkcase-ce-facts.yml
+++ b/vagrant/provisioning/arkcase-ce-facts.yml
@@ -425,7 +425,7 @@ default_database_password: "@rKc@3S!"
 
 # tomcat_major_version needed to build download URL
 tomcat_major_version: 9
-tomcat_version: 9.0.22
+tomcat_version: 9.0.82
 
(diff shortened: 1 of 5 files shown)


david-oc-miller:
Tomcat upgrade by Ansible has been working in the AWS installer branch, as John knows, and I also updated the Pentaho play so we can upgrade the Pentaho Tomcat.  See these two commits: https://github.com/ArkCase/arkcase-ce/commit/8fb0878053be5904bd467558a7e1d8af903db1f2, https://github.com/ArkCase/arkcase-ce/commit/a2c37c0cd270e29695eacc88dc3b6cec0b320a24.  This approach also gives us native openssl in Pentaho, versus Java SSL.

david-oc-miller (approved):
We can deal with Pentaho (Tomcat 8) when the times comes.

PR #586, opened 2023-11-23
Initial commit, need discussion

martinbozinovski:
@david-oc-miller 
This can go in couple of ways. 
1. Everything strip down from the script and executed with as many ansible modules as possible
2. Use it like this

Note*: it is pretty difficult to find the appropriate services that needs restart after all of the changes. We can check for the directory name where changes occurred, and restart the service if corresponded service has that directory name (but this is not a subtle and bulletproof task). If you have some idea, please share it.
Note No.2*: i will leave it in DRAFT till discussion ends.

### DIFF
--- a/vagrant/provisioning/arkcase-all-localhost.yml
+++ b/vagrant/provisioning/arkcase-all-localhost.yml
@@ -123,3 +123,5 @@
       tags: [foia-dev-reports, arkcase-developer]
     - role: arkcase-developer
       tags: [arkcase-developer]
+    - role: upgrade-log4j
+      tags: [upgrade-log4j]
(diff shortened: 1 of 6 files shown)


jshinnick-armedia:
Based on the path, for the Tomcat process, you can scan /usr/lib/systemd/system and /etc/systemd/system using grep -r and the path of your tomcat process. if you get a hit, you can systemctl stop <unit name> to stop the particular tomcat process. if there is not systemctl unit match, then issuing a soft kill to the tomcat process will stop it following by a tomcat/bin/startup.sh executed as the user that owns the tomcat installation.

martinbozinovski:
@jshinnick-armedia 2nd commit is a complete refactor using your proven script. I just sanitize a bit regarding paths, comments, checks before stopping services..

PR #587, opened 2023-12-01
Update main.yml

martinbozinovski:
Bad logic for the when condition

### DIFF
--- a/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
@@ -94,10 +94,10 @@
         name: arkcase
         state: started
       when: _result.status == 200
-  when: solr_data_formated < space_available_formated
+  when: solr_data_formated > space_available_formated
 
 - name: output if there is not space available
   become: yes
   debug:
     msg: "There is no space available under {{ root_folder }}"
-  when:  solr_data_formated > space_available_formated
+  when:  solr_data_formated < space_available_formated
--- a/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
@@ -23,12 +23,8 @@
 
     - name: copy existing solr data to backup folder
       become: yes
-      copy:
-        remote_src: yes
-        src: "{{ root_folder }}/data/solr/"
-        dest: "{{ root_folder }}/data/solr_backup/"
-        owner: solr
-        group: solr
+      become: solr
+      shell: mv {{ root_folder }}/data/solr/ {{ root_folder }}/data/solr_backup/
 
     - name: delete solr cores from previous install
       command: "{{ root_folder }}/app/solr/bin/solr delete -c {{ item }}"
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -94,10 +94,10 @@
         name: arkcase
         state: started
       when: _result.status == 200
-  when: solr_data_formated < space_available_formated
+  when: solr_data_formated > space_available_formated
```
david-oc-miller: Let's say solr_data_formated is 5, and space_available_formated is 10.  Do we need 2x space_available as solr, or 1x?   If we make only one copy, then 1x is fine.  So in this example, 5 < 10 and by the new logic, we wouldn't proceed.  Isn't the old logic right?  Where we would proceed in this example?

martinbozinovski: You have a point. 
If the solr_data_formated is 6 and space_available_formated is 10, is still a valid condition to proceed with that clause. 
I need to think of another solution here. 
Edit: a move instead of a copy (like you suggested before), will solve the problem.

martinbozinovski: @david-oc-miller Please check this latest changes.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -94,10 +90,10 @@
         name: arkcase
         state: started
       when: _result.status == 200
-  when: solr_data_formated < space_available_formated
+  when: solr_data_formated * 2 > space_available_formated
```
david-oc-miller: I still think the operators are reversed.  We want to proceed when the space available is greater than twice the Solr data, and we want to issue a warning and stop when the space available is less than twice the Solr data.  But now it's the other way around.

martinbozinovski: @david-oc-miller You were completely right here, and i was neglecting that because it was working (applied the 1st rule, if it works, do not touch it). And took a deeper dive today on this, and found out that math operations are not interpreted the same on conditionals as in the jinja templating {{ something here }}, ex:  solr_data_formated: "{{ solr_data.stdout | human_to_bytes | int * 2 }}" here we double the integer, and here when: solr_data_formated * 2 > space_available_formated we attach the string solr_data_formated twice and we get solr_data_formatedsolr_data_formated.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -100,4 +96,4 @@
   become: yes
   debug:
     msg: "There is no space available under {{ root_folder }}"
-  when:  solr_data_formated > space_available_formated
\ No newline at end of file
+  when:  solr_data_formated > space_available_formated
```
david-oc-miller, line 99: minor: just for future reference, not to be changed now, `formatted` is spelled with two t's :-)

david-oc-miller (approved):

PR #590, opened 2023-12-05
ACS-1469-Fixes for Solr 8.11.2

satya-udayakumar:


david-oc-miller (approved):

vagrant/provisioning/roles/solr/files/solrconfig.xml
```diff
@@ -35,7 +35,7 @@
        that you fully re-index after changing this setting as it can
        affect both how text is indexed and queried.
   -->
-  <luceneMatchVersion>7.7.2</luceneMatchVersion>
+  <luceneMatchVersion>8.11.2</luceneMatchVersion>
```
martinbozinovski, line 38: If "luceneMatchVersion" is something that can be updated when we change solr_version from facts files, we can use that variable here  and be persistent across solr upgrades

martinbozinovski:
@satya-udayakumar I would argue for the naming of the branches, let it be more descriptive than Solr, something like: acs-1469-fixes-for-solr, or some similar pattern. So, since this is not still merged, please change it.

PR #593, opened 2023-12-13
Add proper module for becoming another user

martinbozinovski:
Well, this is embarrassing :)
But, the code works fine under local use, with ansible-module executed.


vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -23,7 +23,7 @@
 
     - name: copy existing solr data to backup folder
       become: yes
-      become: solr
+      become_user: solr
```
david-oc-miller, line 26: argh how could we miss this one!

david-oc-miller (approved):

PR #594, opened 2023-12-14
Refactor after run on foia test env

martinbozinovski:
Why this refactor?
I've run the code previous against core.test env. It run ok. 
On foia.test we have some differences, like aliases (this should be manually deleted or somehow otherwise managed) for cores. Other mods are regarding speed of the installment itself, different ansible versions (local vs Tower), and proper when checks.

### DIFF
--- a/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
@@ -13,6 +13,18 @@
     space_available_formated: "{{ space_available.stdout | human_to_bytes | int }}"
     solr_data_formated: "{{ solr_data.stdout | human_to_bytes | int * 2 }}"
 
+- name: check if we have some alias for our cores
+  become: yes
+  uri:
+    validate_certs: false
+    url: https://{{ solr_host }}:8983/solr/admin/collections?action=LISTALIASES
+    follow_redirects: none
+  register: core_alias
+
+- name: output info about aliases for core
+  debug:
+    msg: "{{ core_alias.json }}"
+
 - name: upgrade and backup solr if space is available under root_folder
   block: 
     - name: stop arkcase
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -13,6 +13,18 @@
...
 
+- name: check if we have some alias for our cores
+  become: yes
+  uri:
+    validate_certs: false
+    url: https://{{ solr_host }}:8983/solr/admin/collections?action=LISTALIASES
+    follow_redirects: none
+  register: core_alias
+
+- name: output info about aliases for core
```
david-oc-miller, line 24: What do we do with this information?  Does the operator need to remove the aliases before the task proceeds?  Does the operator need to do something to restore these aliases after the upgrade is done?  Or something else?  Can any such work be automated here in Ansible?  In short, why do we need to list the aliases - what action should be taken?

david-oc-miller (approved):

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
martinbozinovski: Yes, the operator need to remove the aliases before deleting the cores because it fails if the cores are part of an alias. I've added that info for simplicity for deleting/renaming the alias before attempting next upgrade. In my opinion, the operator should be aware that the cores are a part of an alias, and should manage appropriate action: delete/rename. Since i don't know what is best for that situation, i just left for the operator to choose. 
Note: can be automated with ansible, but i would argue against it because we don't know the reason for creating the alias. And if the operator knows what he is doing, he would manage some of the above actions and proceed with the upgrade. 
Note No.2: maybe we can improve the code with result from this alias task, and if we have alias, to not proceed with the upgrade at all.

david-oc-miller: Right ... if the subsequent tasks will fail in the presence of extra aliases, then we should output a message with a warning for the operator ("Solr upgrade not completed; please remove the aliases and try again"), and stop.

PR #596, opened 2024-02-22
add changes when we have older standalone solr without collections

martinbozinovski:
@david-oc-miller Please review this latest addition. Hawaii test env had 7.7.2 standalone, not solr_cloud enabled Solr, so it failed on me. If you have some time, please consider some other edge cases to add them.

### DIFF
--- a/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
+++ b/vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
@@ -1,12 +1,17 @@
-- name: Check systemd zookeeper file exists
-  stat:
-    path: /etc/systemd/system/zookeeper.service
-  register: service_file
+- name: Populate service facts
+  service_facts:
 
 - name: install zookeeper
   include_role: 
     name: zookeeper
-  when: not service_file.stat.exists
+  when: ansible_facts.services["zookeeper.service"] is not defined
+
+- name: restart zookeeper if not running
+  become: yes
+  systemd:
+    name: zookeeper
+    state: restarted
+  when: ansible_facts.services["zookeeper.service"].state != "running"
 
 - name: check for available disk space
   become: yes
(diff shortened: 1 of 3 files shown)


vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -1,3 +1,13 @@
+- name: Check systemd zookeeper file exists
```
david-oc-miller: Minor - there's a more idiomatic way to see if a service exists, without depending on the specific location of the service file.  I have this in the AWS installer branch... notice how the `service_facts` is just empty, it just populates standard Ansible facts:

```yaml
- name: check status of fapolicyd service
  service_facts:

- name: some other task that should run when fapolicyd is installed
  ...
  when: ansible_facts.services["fapolicyd.service"] is defined
```

david-oc-miller: ... only small problem is that `service_facts` takes some few seconds to run.

vagrant/provisioning/roles/upgrade-solr/tasks/main.yml
```diff
@@ -1,12 +1,17 @@
...
-    path: /etc/systemd/system/zookeeper.service
-  register: service_file
+- name: Populate service facts
+  service_facts:
 
 - name: install zookeeper
   include_role: 
     name: zookeeper
-  when: not service_file.stat.exists
+  when: ansible_facts.services["zookeeper.service"] is not defined
```
martinbozinovski, line 7: @david-oc-miller You are right, this is better approach. Plus, i added extra steps, just for safety if zookeeper is there, but not running. Since we need him to write configuration for Solr with zkcli.

david-oc-miller (approved):

PR #597, opened 2024-03-05
FIPS AWS - Solr upgrades in a multi-Solr node environment

david-oc-miller:


vagrant/provisioning/roles/solr-replicas-current-node/tasks/replicas.yml
```diff
@@ -0,0 +1,5 @@
+- name: create replicas on current node
+  uri:
+    url: https://{{ solr_host }}:8983/solr/admin/collections?action=ADDREPLICA&collection={{ item }}&node={{ solr_host }}:8983_solr&shard=shard1
+  when: solr_host | default("") != ""
```
martinbozinovski: Can we use something like this:
when: solr_host is defined and solr_host != ""
For functionality wise, it is the same.

david-oc-miller: done

PR #603, opened 2024-04-16
if ecm_viewer is pdftron, snowbound and snowbound-app roles don’t nee…

satya-udayakumar:
…d to be executed.

### DIFF
--- a/vagrant/provisioning/arkcase-all.yml
+++ b/vagrant/provisioning/arkcase-all.yml
@@ -79,8 +79,10 @@
       tags: [core, pentaho, pentaho-configuration, marketplace]
     - role: snowbound
       tags: [core, snowbound, marketplace]
+      when: ecm_viewer is not defined
     - role: snowbound-app
       tags: [core, snowbound, snowbound-app, upgrade, marketplace]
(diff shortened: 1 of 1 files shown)


vagrant/provisioning/arkcase-all.yml
```diff
@@ -79,8 +79,10 @@
       tags: [core, pentaho, pentaho-configuration, marketplace]
     - role: snowbound
       tags: [core, snowbound, marketplace]
+      when: ecm_viewer is not defined
```
martinbozinovski, line 82: This is nice, just need to remember to remove ecm_viewer: "pdftron" from the facts. Because, i can see it in more environments.

martinbozinovski (approved):

vagrant/provisioning/arkcase-all.yml
david-oc-miller: but if it's defined, and set to pdftron, then these roles would be executed; it should be more like:

```yaml
when: (ecm_viewer | default('snowbound')) == 'snowbound'
```

david-oc-miller:
closing per Martin's request

PR #647, opened 2024-09-04
Add working default version for libreOffice package

martinbozinovski:


vagrant/provisioning/roles/arkcase-file-utils/tasks/main.yml
```diff
@@ -71,15 +71,15 @@
...
+      dest: "/tmp/LibreOffice_{{ libreoffice_version | default('24.8.0') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
       owner: root
-      url: "https://ftp.osuosl.org/pub/tdf/libreoffice/stable/{{ libreoffice_version | default('7.3.5') }}/rpm/x86_64/LibreOffice_{{ libreoffice_version | default('7.3.5') }}_Linux_x86-64_rpm.tar.gz --no-check-certificate"
+      url: "https://ftp.osuosl.org/pub/tdf/libreoffice/stable/{{ libreoffice_version | default('24.8.0') }}/rpm/x86_64/LibreOffice_{{ libreoffice_version | default('24.8.0') }}_Linux_x86-64_rpm.tar.gz --no-check-certificate"
   when: not libreoffice_bin_exists.stat.exists
 
 - name: unpack LibreOffice RPM
   become: yes
   unarchive:
-    src: "/tmp/LibreOffice_{{ libreoffice_version | default('7.3.5') }}_Linux_x86-64_install-rpm_en-US.tar.gz"
```
david-oc-miller, line 82: I've been adding libreoffice_version into the facts, so we only have to change the facts, and not the defaults, but this is fine.  It would be better to set one fact (via set_fact) versus repeating the same value 4 times, but we had that before, so it's OK for now.

PR #656, opened 2025-05-13
Ansible change for incoming email

venkatarmedia:


jmcgrady3 (dismissed):

vagrant/provisioning/roles/arkcase-app/templates/server.yml
```diff
@@ -78,6 +78,7 @@ alfresco_port: 443
 email:
   should-delete-messages: false
   should-mark-messages-as-read: true
+  receiver-channel-enabled: {{receiver-channel-enabled | false}}
```
david-oc-miller: Should it be `{{ receiver-channel-enabled | default(false) }}`?

venkatarmedia: yes, addressed the comment

david-oc-miller (approved):

PR #666, opened 2025-06-16
added new set fact task

satya-udayakumar:


fergusonna (approved):

vagrant/provisioning/roles/java-corretto-11/tasks/main.yml
```diff
@@ -42,6 +42,11 @@
 
 #Arkcase
 
+- name: set Java 8 as Java Home (versions below 2023.02)
+  set_fact:
```
david-oc-miller, line 46: At some point it'll make more sense to just add an `arkcase_java_home` fact to the actual facts file.

david-oc-miller (approved):

vagrant/provisioning/roles/java-corretto-11/tasks/main.yml
satya-udayakumar: what if two services use different versions of Java but reference the same variable? thats why we use it like this am I right?

david-oc-miller: hence `arkcase_java_home` so we have possibility of `solr_java_home` etc.

PR #671, opened 2025-07-18
added sudo permissions

satya-udayakumar:


venkatarmedia (dismissed):

vagrant/provisioning/roles/foia/tasks/main.yml
```diff
@@ -22,6 +22,7 @@
   register: html2pdf_exists
 
 - name: Set permissions to 755 on html2pdf_chromium.so
+  become: yes
```
david-oc-miller, line 25: This should work, but would be better to add `become_user: arkcase`

david-oc-miller (approved):

PR #673, opened 2025-09-09
solr-custom-plugin task for installation of the ArkSolrCustomizations Solr plugin

fergusonna:
This task will 

- Enable Solr to load and use Runtime Libraries
- Add the ArkSolrCustomization.jar to Solr's configuration and path
- Update the Solr Schema to replace the Solr PatternTokenizerFactory with one provided by the plugin
- (Optionally) Add JAR signature information to Zookeeper and Solr to prevent unsigned jars from loading

New facts added:
```yaml
solr_plugin_pem_file: "" # the file path to the PEM file used when signing the ArkSolrCustomizations jar, 
# if the ark_solr_custom_jar_sig is defined, but no file is specified,
# a provided PEM file will be used instead. This is what the jar is currently signed. 
ark_solr_custom_jar_sig: "" # the key to provide to Solr when using a signed jar file.
```

As this is related to the Solr Snippets functionality for a specific client, all of this also checks that `enable_solr_snippets = true` before making any changes to Solr.


vagrant/provisioning/arkcase-dev-facts.yml
```diff
@@ -786,4 +786,11 @@ microservice_protocol: http
...
 # Set to true to use the alternative Solr schema, which allows for Solr Highlighting of File Content. 
-enable_solr_snippets: false
\ No newline at end of file
+enable_solr_snippets: false
+ark_solr_customizations_jar_checksum: "43a2e407a4bc75f6e91ad7c06838f572bc7d711e"
+
+# Solr Plugin PEM File requires the path to where the file is stored.
+# IMPORTANT: If a new PEM file is provided, the ArkSolrCustomizations.jar MUST BE RE-SIGNED based on the provided file.
+# See https://solr.apache.org/guide/8_11/adding-custom-plugins-in-solrcloud-mode.html#securing-runtime-libraries for details
+solr_plugin_pem_file: ""
```
david-oc-miller, line 796: Is either `solr_plugin_pem_file` or `ark_solr_custom_jar_sig` required if the jar file is signed?  Or if not specified, is the default pem file used?  (If so specify the default pem in the comments).  If the jar is already signed, does the presence of either fact cause it to be re-signed?  

... in short it's not clear to me how these facts interact with or affect the jar.

vagrant/provisioning/arkcase-dev.yml
```diff
@@ -79,3 +79,5 @@
       tags: [core, privacy, reports, privacy-dev-reports]
     - role: arkcase-developer
       tags: [core, arkcase, arkcase-developer]
+    - role: solr-custom-plugin
```
david-oc-miller, line 82: I believe you can include tags conditionally; like, include the `solr` tag when `enable_solr_snippets` is true.

Example from Ansible docs:

```
- hosts: webservers
  roles:
     - role: debian_stock_config
       when: ansible_facts['os_family'] == 'Debian'
```

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+  become_user: solr
+  file:
+    path: "{{ root_folder }}/app/solr/contrib/arkcase/lib/"
+    state: directory
+  when: "enable_solr_snippets | default(false)"
+
+- name: create public key for signing Solr plugin jar (jar sig must be provided)
+  become: yes
+  command: openssl rsa -in {{ solr_plugin_pem_file | default('/etc/tls/private/arkcase-root.pem') }} -pubout -outform DER -out {{ root_folder }}/common/solr_plugin_key.der
+  when: "(ark_solr_custom_jar_sig is defined and ark_solr_custom_jar_sig != "" and enable_solr_snippets)  | default(false)"
```
david-oc-miller: Nitpick: you can define a variable e.g. `sign_jar` based on these three booleans, and in the remaining conditions just refer to `sign_jar`

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+  wait_for:
+    port: 8983
+    delay: 15
+    timeout: 300
+  when: systemd_solr_updated is changed
+
+- name: add arkSolrCustomizations.jar to the .system blob (required for Solr Cloud mode)
+  ansible.builtin.command:
+    cmd: >
+      curl -X POST -H 'Content-Type: application/octet-stream' --data-binary @roles/files/arkSolrCustomizations.jar --insecure https://{{ internal_host }}:{{ solr_port | default('8983') }}/solr/.system/blob/arkSolrCustomizations
```
david-oc-miller, line 107: Do these system blob changes have to be applied toe each node?

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+  when: "enable_solr_snippets | default(false)"
+
+- name: Confirm arkSolrCustomizations.jar to the .system blob (required for Solr Cloud mode)
+  ansible.builtin.command:
+    cmd: >
+      curl --insecure https://{{ internal_host }}:{{ solr_port | default('8983') }}/solr/.system/blob?omitHeader=true
+  when: "enable_solr_snippets | default(false)"
+
+- name: Add signed arkSolrCustomizations.jar as a runtime lib
```
david-oc-miller: Nitpick: the task name tells me a jar is being uploaded, but the task actually sends a config file.

vagrant/provisioning/arkcase-dev-facts.yml
fergusonna: I've already signed the jar with the default (using [this file](https://github.com/ArkCase/arkcase-ce/blob/develop/vagrant/provisioning/roles/pki-dev/files/tls/private/arkcase-root.pem)). 

The ark_solr_custom_jar_sig is required ONLY if Solr is to only accept a signed jar. If so, then that _sig value needs to match the signature of the jar that's expected. 

If a non-default (not based on arkcase-root.pem) signing is to be used, then both the pem file (it's path as `solr_plugin_pem_file`) and the correct jar signature (as `ark_solr_custom_jar_sig`) are required.

david-oc-miller: This is much more clear to me!  Can you add this comment to the comments?

vagrant/provisioning/arkcase-dev.yml
fergusonna: From my testing, this upgrade needs to be done to an already running Solr instance, since the step that adds the jar as a runtime library (`Add signed arkSolrCustomizations.jar as a runtime lib` or `Add arkSolrCustomizations.jar as a runtime lib without sig value`, if the signature is provide) is adding that runtime library to the configuration of the collection that is to use it.

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
fergusonna: Probably. I'll need to test that on an environment with multiple nodes.

vagrant/provisioning/arkcase-dev-facts.yml
fergusonna: I've updated the comments on the facts. Still need to test the changes that simplify the conditional checks

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
fergusonna: Updated

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
fergusonna: I've added `use_signed_plugin` which checks these 3 and `plugin_pem_file` which uses either the facts-defined file or the default.

david-oc-miller (approved):

vagrant/provisioning/roles/solr-custom-plugin/tasks/main.yml
fergusonna: @david-oc-miller, I only had to run this on node 1, and node 2 reported having the jar available.

fergusonna: I did have to edit this to use the node's hostname, instead of `internal_host` when running the curl commands

PR #697, opened 2026-03-17
ACS-2004: Ensure proper permissions for DOJ Report folder after unzip

srinivasagunturu:


vagrant/provisioning/roles/foia-analytical-reports/tasks/main.yml
```diff
@@ -100,6 +100,16 @@
     src: "{{ root_folder }}/install/pentaho/{{ etl_download_path }}"
     dest: "{{ root_folder }}/install/pentaho/foia-reports-dw{{ foia_analytical_reports_version_formatted }}"
 
+- name: Force correct ownership and group write permissions after unzip
```
david-oc-miller: The unarchive task (right above this one) has `mode`, `owner`, and `group` options.  So unarchive itself should do what you need.

srinivasagunturu: After unarchiving, the group write permission is missing, which causes the installer to fail.
drwxr-sr-x. 2 pentaho pentaho_group     37 Mar 18 08:20 output

david-oc-miller: Is it still missing after you add "group: pentaho_group" to the unarchive task?

srinivasagunturu: Yes, It appears that after unarchiving, we need to reapply the mode to ensure the correct permissions are set.

error_creating_maps.0 - Caused by: java.io.FileNotFoundException: /opt/app/arkcase/install/pentaho/foia-reports-dw-5.0.1-SNAPSHOT/output/error_creating_maps.txt (Permission denied)", "2026/03/18 11:26:16 - tfo: error_creating_maps.0 - \\tat java.base/java.io.FileOutputStream.open0(Native Method)", "2026/03/18 11:26:16 - tfo:

david-oc-miller: Sorry, let me clarify, you mean you did add `group: "pentaho_group"` to the unarchive task, and the unarchived files still did not have pentaho_group?

srinivasagunturu: Update unarchive task to set mode instead of using a separate file task

david-oc-miller (approved):

